PR #4, opened 2019-04-09
Provide JSON structure and individual error IDs to Cryptol errors

kwf:
The structure currently represented only goes one level deep—contained errors (notably parse, renamer, and type errors) are embedded as lists of `pretty`ed strings. In future, this could easily be altered to provide deeper structure—should this PR be merged before then?

### DIFF
--- a/cryptol-remote-api/src/CryptolServer.hs
+++ b/cryptol-remote-api/src/CryptolServer.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE NoMonomorphismRestriction #-}
 module CryptolServer where
 
 import Control.Exception
@@ -7,10 +8,17 @@ import Control.Lens
 import Control.Monad
 import Control.Monad.IO.Class
 import qualified Data.Aeson as JSON
+import Data.Foldable
+import qualified Data.Text as Text
+import qualified Data.Vector as Vector
 
 import Cryptol.Eval.Monad (EvalOpts(..), PPOpts(..))
-import Cryptol.ModuleSystem (ModuleCmd, ModuleEnv, checkExpr, evalExpr, loadModuleByPath, loadModuleByName)
-import Cryptol.ModuleSystem.Env (getLoadedModules, lmCanonicalPath, lmFingerprint, meLoadedModules, initialModuleEnv, meSolverConfig)
+import Cryptol.ModuleSystem
+  (ModuleCmd, ModuleEnv, ModuleError(..), ModuleWarning(..),
+   checkExpr, evalExpr, loadModuleByPath, loadModuleByName)
+import Cryptol.ModuleSystem.Env
+  (getLoadedModules, lmCanonicalPath, lmFingerprint, meLoadedModules,
+   initialModuleEnv, meSolverConfig)
 import Cryptol.ModuleSystem.Fingerprint
 import Cryptol.Parser.AST (ModName)
 import Cryptol.Utils.Logger (quietLogger)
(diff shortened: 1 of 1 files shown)


cryptol-remote-api/src/CryptolServer.hs
```diff
@@ -1,43 +1,43 @@
...
        case out of
          (Left x, warns) ->
-           raise (cryptolError (JSON.toJSON (pretty x, map pretty warns)))
+           raise (cryptolError x warns)
          (Right (x, newEnv), warns) ->
            do setState (set moduleEnv newEnv s)
               return x
 
-data LoadedModule = LoadedModule
+data LoadedModule = LoadedMoodule
```
david-christiansen: This new constructor name seems to be a typo...

cryptol-remote-api/src/CryptolServer.hs
```diff
@@ -77,3 +77,107 @@ validateServerState =
          if fp == Just (lmFingerprint lm)
            then continue
            else return False
+
+--------------------------------------------------------------------------------
+
+-- The schema for translating Cryptol errors/warnings into JSONRPCExceptions
+
+-- Reserved range for Cryptol exceptions: 20,000 - 21,000
+cryptolErrorPrefix :: Integer
```
david-christiansen: Is this a "prefix", or is it more of a "base"?

cryptol-remote-api/src/CryptolServer.hs
```diff
@@ -1,43 +1,43 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE NoMonomorphismRestriction #-}
```
david-christiansen: Why do we need this language extension?

david-christiansen:
> should this PR be merged before then?

I think so. The specific JSON representation of the stuff in the data field on errors is still pretty underspecified and arbitrary in most of these, and this current change makes it more specified and less arbitrary.

cryptol-remote-api/src/CryptolServer.hs
kwf: Mooooodule 🐮 

```suggestion
data LoadedModule = LoadedModule
```

cryptol-remote-api/src/CryptolServer.hs
kwf: These helper functions have polymorphic types which will be inferred as monomorphic otherwise:

```haskell
    jsonString = JSON.String . Text.pack
    jsonPretty = jsonString . pretty
    jsonShow = jsonString . show
    jsonList = JSON.Array . Vector.fromList
```

The alternative is giving them all signatures, which I deemed unnecessary in this case.

kwf:
> I think so. The specific JSON representation of the stuff in the data field on errors is still pretty underspecified and arbitrary in most of these, and this current change makes it more specified and less arbitrary.

Sounds good to me. We'll just have to keep in mind that changes to this bit will constitute breaking API changes—but those will be easy enough to adapt to, and ideally we'll soon have tests that exercise this chunk of the API so we can see if we break it.

kwf:
Another thought: The way Cryptol represents errors means that a single call can produce multiple errors, with potentially different types. Specifically, you can get something like a `RenamerErrors` which contains a list of different `RenamerError`s. This means we can't consider individual sub-errors as having a specific JSON-RPC error number, because there could be several. If we care about machine parsing for these lists of structured sub-errors, we'll need to devise our own schema for how to communicate which one(s) occurred.

david-christiansen:
Yes, I think it's a big premature to worry about breaking API changes, as the API is as yet mostly nonexistent. But eventually, everything in JSON becomes part of the documented API, which is why I've been trying to only define ToJSON/FromJSON instances for newtype wrappers or for datatypes that exist only as a representation of the protocol.

cryptol-remote-api/src/CryptolServer.hs
david-christiansen: Won't type signatures help people who are reading the code?

kwf:
> But eventually, everything in JSON becomes part of the documented API, which is why I've been trying to only define ToJSON/FromJSON instances for newtype wrappers or for datatypes that exist only as a representation of the protocol.

I agree—that's why I didn't make a `ToJSON` instance for cryptol errors, instead just making it a function (also, it would be an orphan instance if I had done that).

cryptol-remote-api/src/CryptolServer.hs
kwf: In this case I thought these functions were so small that the type signatures wouldn't add much; if you felt they'd have helped you, then that's enough evidence for me to add them of course!

david-christiansen:
Were we going to go with the language extension, or with type signatures?

cryptol-remote-api/src/CryptolServer.hs
kwf: I'm fine with "base"—changed in 71a7d1d.

cryptol-remote-api/src/CryptolServer.hs
kwf: Resolved in 8ba57a0 in favor of using explicit type signatures.

kwf:
All fixed up, sorry for the late merge!

david-christiansen:
Thanks for doing it!

Den tir. 23. apr. 2019 kl. 13.23 skrev Kenny Foner <notifications@github.com

PR #5, opened 2019-04-10
Add standardized debug logging

david-christiansen:
Now, methods have an effect available to use for printf-debugging.

### DIFF
--- a/argo/src/Argo/Socket.hs
+++ b/argo/src/Argo/Socket.hs
@@ -7,8 +7,11 @@ module Argo.Socket
 import Control.Concurrent       (forkFinally)
 import Control.Concurrent.Async (Async, async, forConcurrently_)
 import Control.Exception        (displayException)
+import Control.Lens             (view)
 import Control.Monad            (forever)
-import System.IO                (Handle, IOMode(ReadWriteMode), hClose, hPutStrLn)
+import Data.String              (IsString(..))
+import qualified Data.Text as T
+import System.IO                (Handle, IOMode(ReadWriteMode), hClose, hPutStrLn, stderr)
 import qualified Data.Map as Map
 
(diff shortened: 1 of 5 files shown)


david-christiansen:
In particular, this beats out `Debug.Trace` because we can arrange for it to go to the right place, which can differ based on how the app is served.

argo/src/Argo/Socket.hs
```diff
@@ -84,15 +85,15 @@ startListening addr =
...
+acceptClient :: App s -> N.Socket -> IO ()
+acceptClient app s =
 
   do (c, peer) <- N.accept s
      h         <- N.socketToHandle c ReadWriteMode
      -- don't use c after this, it is owned by h
 
      log ("CONNECT: " ++ show peer)
-     forkFinally (serveHandlesNS logH h h app) $ \res ->
+     forkFinally (serveHandlesNS (Just stderr) h h app) $ \res ->
```
kwf: Do we want to hard-code the use of `stderr` here? For now it seems fine, and perhaps the simplicity of making the decision in this location warrants the (slight) loss of generality as compared to letting the caller specify the handle. What do you think?

kwf:
Except for the question above, everything here looks good to me—this is how I would have implemented this feature as well.

argo/src/Argo/Socket.hs
david-christiansen: I thought about it the way you did. I went with simplicity when I couldn't figure out any situation where I'd want something else. If such a situation arises, we can always add the parameter.

david-christiansen:
Thanks! Mind if I merge, given that answer?

kwf:
Sounds good to me!

kwf:
(and so it is done)

PR #9, opened 2019-07-10
Protocol docs

david-christiansen:
When you have time, I'd appreciate a review from the perspective of another implementer of a client, @kwf. Thanks!


david-christiansen:
CC @atomb

kwf:
6c90c89 was exactly what I had in mind :)

PR #20, opened 2019-09-28
Working but flawed execution of Python bindings from Haskell tests (WIP)

david-christiansen:
The problem here is that the organization of the data-files field in
the Cabal file seems fragile and subject to maintenance pain (in
particular, listing all the Python files appropriately is difficult to
test unless CI does an sdist first).

This does seem to be a useful starting point, however.


kwf:
According to [this documentation](https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-data-files), `**` recursive glob wildcards are permissible in `data-files` specifications since Cabal 2.4, with the following caveat:

> On efficiency: if you use ** patterns, the directory tree will be walked starting with the parent directory of the **. If that’s the root of the project, this might include .git/, dist-newstyle/, or other large directories! To avoid this behaviour, put the files that wildcards will match against in their own folder.

In this case, what do you think about making a `python` directory (to skirt the above issue) and setting `data-files: python/**.py` plus `requirements.txt`? That way all python files will be scooped up without needing to change anything about the cabal file if we add new directory structures.

david-christiansen:
I think, actually, that 7be660c addressed the issue of forgetting to include Python files, but the dummy package should probably also use the `**` feature.

kwf:
When I run `cabal new-test saw-remote-api` or `cabal new-test cryptol-remote-api` I get the following error (with random temp directory names of course):

```
pip failed in environment \"/private/var/folders/3s/jkj9nzz96qq18ttz4qghk1l80000gp/T/virtenv-8af919cc0275cb02\" with code 1:
stdout:
stderr: Directory '/Users/kwf/Work/protocol/python/.' is not installable. Neither 'setup.py' nor 'pyproject.toml' found.
You are using pip version 19.0.3, however version 19.2.3 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

Is this working on your machine? It seems like it shouldn't be trying to install anything to the *actual* `/Users/kwf/Work/protocol/python/` directory, but rather to some temp-directory mirror thereof?

david-christiansen:
Looks like I forgot to commit a file. Try again!

kwf:
That fixes the above issue, but there's still a test failure for `cryptol-api_test.py`:

```
Traceback (most recent call last):
          File "/Users/kwf/Work/protocol/cryptol-remote-api/test-scripts/cryptol-api_test.py", line 21, in <module>
            cryptol.add_cryptol_module('Foo', c)
          File "/private/var/folders/3s/jkj9nzz96qq18ttz4qghk1l80000gp/T/virtenv-092e425326c8f37e/lib/python3.7/site-packages/cryptol/__init__
.py", line 337, in add_cryptol_module                                                                                                        
            sys.modules[name] = CryptolModule(connection)
          File "/private/var/folders/3s/jkj9nzz96qq18ttz4qghk1l80000gp/T/virtenv-092e425326c8f37e/lib/python3.7/site-packages/cryptol/__init__
.py", line 322, in __init__                                                                                                                  
            cryptoltypes.to_schema(x['type']),
          File "/private/var/folders/3s/jkj9nzz96qq18ttz4qghk1l80000gp/T/virtenv-092e425326c8f37e/lib/python3.7/site-packages/cryptol/cryptolt
ypes.py", line 320, in to_schema                                                                                                             
            [to_prop(p) for p in obj['propositions']],
          File "/private/var/folders/3s/jkj9nzz96qq18ttz4qghk1l80000gp/T/virtenv-092e425326c8f37e/lib/python3.7/site-packages/cryptol/cryptolt
ypes.py", line 320, in <listcomp>                                                                                                            
            [to_prop(p) for p in obj['propositions']],
          File "/private/var/folders/3s/jkj9nzz96qq18ttz4qghk1l80000gp/T/virtenv-092e425326c8f37e/lib/python3.7/site-packages/cryptol/cryptolt
ypes.py", line 335, in to_prop                                                                                                               
            return Logic(to_type(obj['subject']))
        KeyError: 'subject'
```

david-christiansen:
That's very odd. It works locally for me.

How exactly are you invoking the test suite for `cryptol-remote-api`? I'm doing it with `cabal new-test saw-remote-api cryptol-remote-api argo`.

kwf:
That exact command yields the above error :/

david-christiansen:
I was having a similar error before, because `cryptol-remote-api` was putting trailing spaces in some JSON key names. Is it possible that you're running an older version of the Haskell code somehow?

david-christiansen:
I just did a clean and rebuild, and they're still passing.

However, I noticed that when I ran the test command, it didn't rebuild everything. This makes me think that the test suite won't provoke the build of the executables, necessarily. Can you try one more time, first without and then with a `cabal new-build all`?

atomb:
I get a similar error:

````
You are using pip version 19.0.3, however version 19.2.3 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Directory '/Users/atomb/galois/argo/saw-remote-api/../python' is not installable. Neither 'setup.py' nor 'pyproject.toml' found.
You are using pip version 19.0.3, however version 19.2.3 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
````

After that, the tests fail because they can't find the `argo` package, but I think that's because of the earlier bit. Interestingly, `pip --version` says it's version 19.2.3.

david-christiansen:
@atomb Does the version of the branch you're testing include commit 3bf9f8a ? It should have addressed that...

david-christiansen:
I talked to @atomb and with the latest version of this branch, tests pass on his machine. I'm going to merge, on the assumption that @kwf's computer has a configuration issue.

PR #40, opened 2019-11-24
Cleaner process cleanup

### LINKED ISSUE #32: Python does not shut down server process (macOS only?)
On macOS, under some circumstances yet to be diagnosed, Python does not shut down the server process, instead leaving it running indefinitely. This behavior has yet to be replicated on Linux. What's up with that? Let's figure out!

kwf:
Fixes #32. This commit enables `with` blocks for all connection types in the Python API for more fine-grained and principled resource management, and also resolves the failure-to-kill issue observed on macOS.

- [x] **Needs review by a Linux user**: Does the process-killing and management still work on Linux? We should double-check before merging (I only have a Mac at hand presently).
- [ ] Before merging, let's add a regression test to make sure the process-killing failure doesn't emerge again (or on another system). Question to self: where should this live?

### DIFF
--- a/cryptol-remote-api/test-scripts/cryptol-api_test.py
+++ b/cryptol-remote-api/test-scripts/cryptol-api_test.py
@@ -6,18 +6,18 @@ import cryptol.cryptoltypes
 
 dir_path = os.path.dirname(os.path.realpath(__file__))
 
-c = cryptol.connect("cabal new-exec --verbose=0 cryptol-remote-api -- --dynamic4")
+with cryptol.connect("cabal new-exec --verbose=0 cryptol-remote-api -- --dynamic4") as c:
 
-c.change_directory(dir_path)
+    c.change_directory(dir_path)
 
-c.load_file("Foo.cry")
+    c.load_file("Foo.cry")
 
-val = c.evaluate_expression("x").result()
+    val = c.evaluate_expression("x").result()
 
-assert c.call('add', b'\0', b'\1').result() == b'\x01'
-assert c.call('add', bytes.fromhex('ff'), bytes.fromhex('03')).result() == bytes.fromhex('02')
+    assert c.call('add', b'\0', b'\1').result() == b'\x01'
+    assert c.call('add', bytes.fromhex('ff'), bytes.fromhex('03')).result() == bytes.fromhex('02')
 
 
-cryptol.add_cryptol_module('Foo', c)
-from Foo import *
-assert add(b'\2', 2) == b'\4'
+    cryptol.add_cryptol_module('Foo', c)
+    from Foo import *
+    assert add(b'\2', 2) == b'\4'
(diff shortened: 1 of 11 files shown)


david-christiansen:
The test suite passes on Fedora, without me seeing any zombie processes or anything like that.

david-christiansen:
While it's hard to argue with success, I'm not entirely sure about the `with` block setup. I have a few questions that I think help shed light on it:

 * What are the semantics of snapshots with respect to a with block? When control leaves a `with`, the resource is invalidated. Does this mean that snapshots are also invalidated?

 * Does the fix to the Mac process issue allow the `with`-less test scripts to work? Or do `with` blocks and the change to the kill call used depend on one another?

 * Does this change REPL interactions with a SAW server?

Background: I'm a bit hesitant to require the whole program to be wrapped in a `with` block because I think it brings the management of the server as a resource front and center - the whole program is literally subordinated to it. I think the system will be more pleasant to use if the presence of the server is only something that people need to think about in more advanced use cases, like when they want to control caching to make proofs go through faster. If I had my druthers, there'd be no explicit connection step at all, though I don't know how to make that nice to use yet.

Also, this is orthogonal, but Travis is now complaining about Python setup rather than timing out. The package it wants is already schedule to be installed. Do you know whether this was introduced in this branch, or whether it's due to something else?

kwf:
The `with` block setup and the more effective process killing are totally orthogonal: I've tested scripts in the original style, and they work just fine and appropriately clean up resources. (As a result this shouldn't change REPL interactions.) I introduced the `with` functionality with the idea that fine grained resource lifetime management would be a good idea to have as an option, but on reflection the next day, I'm not sure I can come up with many ordinary use cases that would necessitate this, even more so in the future when the server process is a semi-persistent entity.

Snapshots: since the process is killed when the connection goes out of scope in the `with` block, that does indeed mean a snapshot would leak a closed connection if it escaped that scope! Good catch—that in itself might be reason enough to revert the `with` block addition.

> Background: I'm a bit hesitant to require the whole program to be wrapped in a `with` block because I think it brings the management of the server as a resource front and center - the whole program is literally subordinated to it. I think the system will be more pleasant to use if the presence of the server is only something that people need to think about in more advanced use cases, like when they want to control caching to make proofs go through faster. If I had my druthers, there'd be no explicit connection step at all, though I don't know how to make that nice to use yet.

Yup, that agrees with my own aesthetic, thanks for bringing that to the fore. The ideal user experience would be identical to using a native python library, more or less, I think?

Re: Travis: It works on my machine™—I'll look into why not for poor Travis.

kwf:
I've reverted the `with`-block changes and confirmed that the tests pass on my machine. I'll wait for Travis to, uhhh, time out, before merging.

david-christiansen:
LGTM :-)

david-christiansen:
(assuming the tests all pass for you)

Issue #28, opened 2019-11-11
Proof dashboard API


kwf:
Use a logging framework or an HTML generation scheme (or both) to produce a nice report from a proof script of the things that were verified. Python reflection will let us grab the names of spec classes, etc.

david-christiansen:
This seems to presently be more working than this ticket specifies. In particular, we have a dynamically-generated dashboard!

As part of the dashboard work, I'd like to see if we can factor it out into an independent library that anyone could have developed, in principle. This will help us maintain it, and it will also expose which primitives are necessary as part of the public SAW Python API for third-party extension.

david-christiansen:
I imagine we can get by with something like the hook mechanism in Emacs, where libraries can register callbacks that are triggered on certain events in the library, or alternatively we can use the standard Python logging interface as it says above.

kwf:
This is a larger ongoing piece of work which will acquire its own more specific tickets—one single tracking ticket isn't useful at this point, so we're closing this issue and creating issues for the more specific technical work.

PR #41, opened 2019-11-25
WIP: Proof dashboard

kwf:
This is a work-in-progress to support a "proof dashboard" to address #28. Currently, an extension of the `SAWConnection` API enables the following style of proof:

```python
c = saw.connect("cabal new-exec --verbose=0 saw-remote-api -- --dynamic4",
                delay_verification_exceptions = True)
...
c.llvm_verify(...).result()
...
c.assert_all_verifications()
```

When the script is run, if any of the `.llvm_verify` commands in that session failed, the final line will throw an `AssertionError` with a pretty-printed message describing all the verifications in the file.

This pretty-printed list of verifications (which includes both valid and invalid ones) can be accessed using the `.readable_summary()` method of a `SAWConnection`; a structured log can be gotten using the `.summary()` method.

Note when using this that there is "magic" Pythonic introspection happening: we get filename and line number information for the calls to `llvm_verify` "out of thin air." This functionality may be something we want to delegate to a higher-level API before merging—specifically, the yet-to-be-created "synchronous" API which automatically inserts calls to `.result()`.

### DIFF
--- a/python/argo/interaction.py
+++ b/python/argo/interaction.py
@@ -86,7 +86,6 @@ class Command(Interaction):
     corresponding command's appropriate representation.
     """
 
-
     def _result_and_state(self) -> Tuple[Any, Any]:
         res = self.raw_result()
         if 'error' in res:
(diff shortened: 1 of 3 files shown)


kwf:
Thoughts from discussion:

* We should use a standard logging library, i.e. https://docs.python.org/2/library/logging.html
* It might make more sense for the verification-error-as-logged-boolean behavior to exist at a higher level of the API. This level might do something like, "if the verification failed, log it and assume counterfactually that the lemma was true."

pnwamk:
Archiving this PR. To whom it may concern, if there is renewed action/interest/etc please feel free reopen it =)

PR #48, opened 2020-03-26
Add a --public flag to bind a public socket

david-christiansen:
While doing this, the substantially similar main actions in both servers were consolidated into one, reducing duplication, and the usage information blurb was greatly improved.


david-christiansen:
The tests pass locally, so I'm going to go ahead and merge this.

Issue #62, opened 2020-04-08
Bump dependencies, April 2020


david-christiansen:
We should be using the latest mypy, SAW, and Cryptol (and associated subcomponents).

david-christiansen:
There have been changes to the Cryptol evaluator recently that make this task a bit more difficult on the Haskell side. But it's underway.

david-christiansen:
Still awaiting #63

PR #63, opened 2020-04-08
Update Python code to newest mypy release

david-christiansen:
This is part of addressing #62.

@kwf, can you confirm that I understood the dashboard code I had to change?


david-christiansen:
@kwf said it was good to go.

PR #70, opened 2020-04-21
Feature/saw track files

### LINKED ISSUE #27: Invalidate SAW server cache due to file modification
The Cryptol server uses the `validateServerState` command to invalidate server states when a file changes on disk: https://github.com/GaloisInc/argo/blob/7cb77e1db2b67d5d783cf6def4eb6f9f0242310e/cryptol-remote-api/src/CryptolServer.hs#L73-L83 We should do this with the SAW server too, for LLVM modules, and anything else that touches the disk. This may require altering the set of things that is stored in the server state.

kwf:
This fixes #27 by implementing file hash tracking for LLVM modules loaded, which can be extended to any file we choose to manipulate on disk in the future.

Unlike in Cryptol, loaded LLVM modules in SAW are not associated with a file fingerprint when they are loaded, so we supply this checksum functionality storing a SHA256 hash in an added `Map FilePath (Digest SHA256)` in the `SAWState` for the server.

### DIFF
--- a/saw-remote-api/src/SAWServer.hs
+++ b/saw-remote-api/src/SAWServer.hs
@@ -172,6 +172,12 @@ initialState =
                 }
      return (SAWState emptyEnv bic [] ro rw M.empty)
 
+-- NOTE: KWF: 2020-04-22: This function could introduce a race condition: if a
+-- file changes on disk after its hash is computed by validateSAWState, but
+-- before the function returns and its result is used to inform whether to
+-- recompute a cached result, the cached result may be used even if it is
+-- associated with stale filesystem state. See the discussion of this issue at:
+-- https://github.com/GaloisInc/argo/pull/70#discussion_r412462908
 validateSAWState :: SAWState -> IO Bool
 validateSAWState sawState =
   checkAll
--- a/saw-remote-api/src/SAWServer.hs
+++ b/saw-remote-api/src/SAWServer.hs
@@ -20,6 +20,7 @@ import Data.Parameterized.Some
 import Data.Text (Text)
 import qualified Data.Text as T
 import qualified Crypto.Hash as Hash
+import qualified Crypto.Hash.Conduit as Hash
 
 import qualified Cryptol.Parser.AST as P
 import qualified Cryptol.TypeCheck.AST as Cryptol (Schema)
@@ -175,6 +176,7 @@ validateSAWState :: SAWState -> IO Bool
 validateSAWState sawState =
   checkAll
     [ CryptolServer.validateServerState cryptolState
+    , checkAll $ uncurry checkHash <$> M.assocs (view trackedFiles sawState)
     ]
   where
     checkAll [] = pure True
@@ -184,10 +186,14 @@ validateSAWState sawState =
            then checkAll cs
            else pure False
 
+    checkHash path hash =
+      do currentHash <- Hash.hashFile path
+         pure (currentHash == hash)
+
     cryptolState =
-      CryptolServer.ServerState Nothing moduleEnv
-    moduleEnv =
-      CryptolEnv.eModuleEnv . rwCryptol . view sawTopLevelRW $ sawState
+      CryptolServer.ServerState Nothing
+        (CryptolEnv.eModuleEnv . rwCryptol . view sawTopLevelRW $ sawState)
+
 
 newtype SAWEnv =
   SAWEnv { sawEnvBindings :: Map ServerName ServerVal }
--- a/saw-remote-api/src/SAWServer.hs
+++ b/saw-remote-api/src/SAWServer.hs
@@ -19,6 +19,7 @@ import Data.Parameterized.Pair
 import Data.Parameterized.Some
 import Data.Text (Text)
 import qualified Data.Text as T
+import qualified Crypto.Hash as Hash
 
 import qualified Cryptol.Parser.AST as P
 import qualified Cryptol.TypeCheck.AST as Cryptol (Schema)
@@ -28,6 +29,7 @@ import qualified Lang.Crucible.JVM as CJ
 import Text.LLVM.AST (Type)
 import qualified Verifier.Java.Codebase as JSS
 import Verifier.SAW.CryptolEnv (CryptolEnv)
+import qualified Verifier.SAW.CryptolEnv as CryptolEnv
 import Verifier.SAW.Module (emptyModule)
 import Verifier.SAW.SharedTerm (mkSharedContext, scLoadModule, Term)
 import Verifier.SAW.Term.Functor (mkModuleName)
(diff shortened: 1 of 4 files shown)


saw-remote-api/src/SAWServer/LLVMCrucibleSetup.hs
```diff
@@ -237,4 +238,5 @@ llvmLoadModule (LLVMLoadModuleParams serverName fileName) =
               Left err -> raise (cantLoadLLVMModule (LLVM.formatError err))
               Right llvmMod ->
                 do setServerVal serverName llvmMod
+                   trackFile fileName
```
david-christiansen, line 241: There is a race condition here. I think that avoiding it will entail changes to SAW similar to this change to Cryptol: https://github.com/GaloisInc/cryptol/pull/576

Basically, the file could change between the invocation of `llvmLoadModule` and `trackFile`. Ideally, we get a hash back from the module loading code in addition to the parsed result.

saw-remote-api/src/SAWServer/TrackFile.hs
```diff
@@ -0,0 +1,24 @@
...
+-- @forgetFile@ is used to remove tracking of this file.
+trackFile :: FilePath -> Method SAWState ()
+trackFile path =
+  do hash <- Hash.hashFile path
+     modifyState $ over trackedFiles (M.insert path hash)
+
+-- | Stop tracking a given file. Any state that causally descends from the
+-- moment this method is invoked will not be invalidated by changes to the file
+-- on disk, even if this file was previously tracked via @trackFile@.
+forgetFile :: FilePath -> Method SAWState ()
```
david-christiansen, line 22: In what circumstances might this be useful? AFAIK we don't have a way to unload a module.

saw-remote-api/src/SAWServer.hs
```diff
@@ -162,10 +170,29 @@ initialState =
...
                 }
-     return (SAWState emptyEnv bic [] ro rw)
+     return (SAWState emptyEnv bic [] ro rw M.empty)
 
 validateSAWState :: SAWState -> IO Bool
-validateSAWState _ = pure True
+validateSAWState sawState =
+  checkAll
+    [ CryptolServer.validateServerState cryptolState
+    , checkAll $ uncurry checkHash <$> M.assocs (view trackedFiles sawState)
```
david-christiansen, line 179: Given that we're working in `IO`, but `<$>` is happening on `[]`, it might be easier to read if `map` were written here. Or even a list comprehension. But this is a quite minor style point!

saw-remote-api/src/SAWServer/LLVMCrucibleSetup.hs
kwf: As an intermediary measure, could we hash the file both before and after, and fail if any of the three total hashes mismatch?

saw-remote-api/src/SAWServer/TrackFile.hs
kwf: So far, I didn't have any use case in mind save a hypothetical "unload" command—I included it because it was easy to write.

saw-remote-api/src/SAWServer.hs
kwf: Fixed by https://github.com/GaloisInc/argo/pull/70/commits/b61eb9ac2f1e790117f74931eaa8731cdc0d1352

saw-remote-api/src/SAWServer/LLVMCrucibleSetup.hs
kwf: Ah, I see—I just looked at the implementation in Cryptol and I now understand why the above double-fingerprint solution would be inadequate too: if the file is changed between the first fingerprinting and the module load, and then changed *back* to its original state before the second fingerprinting, there could again be a mismatch. I'm quite hesitant to merge a potential race condition to caching code. Seems like we should implement the same functionality in SAW's LLVM modules... I'll look into what that would entail.

kwf: The LLVM verifier [loads modules from disk like so](https://github.com/GaloisInc/llvm-verifier/blob/af5f873253db889dc7e22c172894f853c52a4218/src/Verifier/LLVM/Codebase.hs#L207-L214):

```haskell
-- | Load a module, calling fail if parsing or loading fails.
loadModule :: FilePath -> IO L.Module
loadModule bcFile = do
  eab <- CE.mapException (LLVMLoadMiscException bcFile) $
         BC.parseBitCodeFromFile bcFile
  case eab of
    Left msg  -> CE.throw (LLVMLoadParseFailure bcFile msg)
    Right mdl -> return mdl
```

In order to prevent the race condition, we would need to replace the call to `BC.parseBitCodeFromFile` (which is [`Data.LLVM.BitCode.parseBitCodeFromFile`](https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/master/src/Data/LLVM/BitCode.hs#L24-L25)) with a simultaneous load/hash. Currently, `BC.parseBitCodeFromFile` is defined (in [`llvm-pretty-bc-parser`](https://github.com/GaloisInc/llvm-pretty-bc-parser)) as (some imports elided):

```haskell
import Data.LLVM.BitCode.Bitstream (Bitstream, parseBitCodeBitstream)
import qualified Data.ByteString as S

parseBitCodeFromFile :: FilePath -> IO (Either Error Module)
parseBitCodeFromFile  = parseBitCode <=< S.readFile

parseBitCode :: S.ByteString -> IO (Either Error Module)
parseBitCode  = parseBitstream . parseBitCodeBitstream

parseBitstream :: Either String Bitstream -> IO (Either Error Module)
parseBitstream e = -- ... (elided)
```

The simplest solution would be to load the whole file as a `ByteString` into memory within SAW, then hash the ByteString, then parse it into an LLVM `Module`. This would suffice to ensure that the loaded `Module` corresponds to the file contents with the hash we compute.

**A note on a possible optimization:** This would not have worse memory usage than the present situation. However, this does bring to light a possible memory usage optimization: we may not need to load the `ByteString` strictly into memory to begin with: we could stream it through both the hash function and the parser without allocating the intermediate strict `ByteString`. This refactor is not necessary to implement this change, however—I'm noting it here merely for the purpose of pointing out its possibility.

kwf: After discussion with @david-christiansen, we've realized that this solution does not in fact remove the race condition—and indeed, there is no way to prevent such a race condition without some kind of transactional file-locking mechanism, because a file could always be changed on disk during the hash-*checking* step, after its own hash has been computed but before that hash has been compared to its previous hash. This possibility exists in the Cryptol module reloading code also, with similar caveats. As such we're punting on avoiding this race condition, and merging with the knowledge that it exists.

PR #72, opened 2020-04-30
Feature/persistent server

### LINKED ISSUE #35: Start server as necessary and maintain it for future runs
The present Python API doesn't have a good way to use an existing server - each run creates a new process and then destroys it at the end. 

It should be possible to start servers if not running, and re-use existing servers if they are. This would allow state to be saved from one run to the next, saving time.

### LINKED ISSUE #58: Extract proof dashboard to independent library
The proof dashboard should be refactored to be a library that could have been written by a third party. This way, we'll discover which primitives we need to expose and how to do it.

kwf:
This PR implements persistence of the server process, with the ability to name a given session so that future invocations of the server pointed at the same session name will report the port for that session rather than instantiating a new server process. (That is, fixes #35.)

Now, having been merged with a branch for detangling the dashboard code, this PR also fixes #58.

Items included in this PR:

- [x] the Argo library now supports a `--session` argument to specify a named session to create or look up (atomicity enforced using file locks)
- [x] docs need to be updated, since the command line interface has changed
- [x] cleanup of the temp directory tracking sessions needs to happen, or else there will be a (albeit glacially slow) filesystem space leak
- [x] double-check that there are no outlying examples outside the test suite that invoke the old command line flags
- [x] change Python library to use a default overrideable session name, and alter its process management so that the server process outlives the Python script

### DIFF
--- a/python/argo/connection.py
+++ b/python/argo/connection.py
@@ -6,7 +6,7 @@ import re
 import socket
 import subprocess
 import signal
-import sys # pylint: disable=unused-import
+import sys  # pylint: disable=unused-import
 from typing import Any, Dict, List, Mapping, Optional, Union
 from mypy_extensions import TypedDict
 
@@ -14,10 +14,10 @@ from . import netstring
 
 
 # Must be boxed separately to enable sharing of connections
-class IDSource: # pylint: disable=too-few-public-methods
+class IDSource:  # pylint: disable=too-few-public-methods
     """A source of unique identifiers for JSON RPC requests."""
 
-    next_id : int
+    next_id: int
 
     def __init__(self) -> None:
         self.next_id = 0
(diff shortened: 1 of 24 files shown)


python/argo/connection.py
```diff
@@ -73,15 +78,23 @@ def setup(self) -> None:
...
+        else:
+            return None
+
     def __del__(self) -> None:
         if self.proc is not None:
-            try:
-                os.killpg(os.getpgid(self.proc.pid), signal.SIGKILL)
-            except ProcessLookupError:
-                pass
+            if not self.persist:
```
kwf, line 93: This is the important bit: when the process handle is GC'ed, we don't kill the underlying process if `self.persist` is set to True.

PR #73, opened 2020-04-30
Allow JSON null for the initial state

### LINKED ISSUE #54: Initial state field: null?
We have a bit of an incongruity in our protocol.

On the one hand, we say that the concrete structure of the `state` fields is intended to be abstract, and that clients should not construct them themselves. On the other hand, the initial state is clearly an array `[]`, and the subsequent states sure look like the kind of thing a user might build.

I propose that we replace the initial state with the JSON null value instead. We could additionally encode states through a `Binary` instance and base64 them on the wire to make it clear to users that they are an implementation detail that is subject to change.

david-christiansen:
We have not yet ruled out [] so as to not break client code.

Fixes #54.


kwf:
Looks all good to me!

Issue #59, opened 2020-04-08
More informative error messages from SAW


david-christiansen:
Right now, when a proof fails, it just says "Proof failed". There should be a structured representation of _why_ the proof failed, returned to the client for inspection.

This issue is likely to require lots of small changes and improvements, both here and in SAW itself - please consider it more of a tracking issue for the topic.

PR #77, opened 2020-05-20
Modify Argo protocol and Python SAW client to track stdout/err

kwf:
Working towards addressing #59 (but not resolving it)...

The Argo protocol now transmits the full captured stdout/err of the
method invoked back to the client, which may do with it what they
please. In the case of the Python client to the SAW server, this
information is exposed in the exceptions thrown, and in particular, to
the View implementations that can track and display this information as
if it was coming natively from the Python process.

To use this functionality in the Python client, add a `View` for it, as the salsa20 and swap examples do:

```python
connect("cabal -v0 v2-run exe:saw-remote-api")
view(DebugLog(err=None))
view(LogResults())
```

The two optional named parameters for `DebugLog` are `out` and `err` and direct the stdout/stderr of the invoked method to either the same output pipe (default), a given file handle, or into the void if specified as `None`.

Because the order of printing is dependent on the order you attach the `View`s, it's usually best to attach the debug-log and results-log in the order above, because that means you see the debug statements prior to the final result, as opposed to the other way.

This PR does not yet include documentation of this feature or protocol change, and we may want to add such before merging.

### DIFF
--- a/argo/src/Argo.hs
+++ b/argo/src/Argo.hs
@@ -9,6 +9,8 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE RankNTypes #-}
 {-# OPTIONS_GHC -Wno-redundant-constraints -Wno-name-shadowing #-}
 
 -- | An implementation of the basic primitives of JSON-RPC 2.0.
@@ -184,6 +186,10 @@ data JSONRPCException =
     , errorID :: Maybe RequestID
       -- ^ The request ID, if one is available. @Nothing@ omits
       -- it, because JSON-RPC defines a meaning for null.
+    , errorStdOut :: String
+      -- ^ The standard output of the method producing this error
+    , errorStdErr :: String
+      -- ^ The standard error of the method producing this error
     } deriving Show
 
 instance Exception JSONRPCException
@@ -198,8 +204,12 @@ instance JSON.ToJSON JSONRPCException where
           Just theID -> JSON.toJSON theID
       , "error" .=
         (JSON.object $ ["code" .= JSON.Number (fromInteger (errorCode exn))
-                       , "message" .= JSON.String (message exn)] ++
-                       (("data" .=) <$> maybeToList (errorData exn)))
+                      , "message" .= JSON.String (message exn)
+                      , "data" .= (JSON.object $
+                        ([ "stdout" .= errorStdOut exn
+                         , "stderr" .= errorStdErr exn ] ++
+                         (("data" .=) <$> maybeToList (errorData exn))))
+                      ])
       ]
 
 -- | A method was provided with incorrect parameters (from the JSON-RPC spec)
@@ -209,6 +219,8 @@ invalidParams msg params =
                    , message = T.pack ("Invalid params: " ++ msg)
                    , errorData = Just params
                    , errorID = Nothing
+                   , errorStdOut = ""
+                   , errorStdErr = ""
                    }
 
 -- | The input string could not be parsed as JSON (from the JSON-RPC spec)
@@ -218,6 +230,8 @@ parseError msg =
                    , message   = "Parse error"
                    , errorData = Just (JSON.String msg)
                    , errorID   = Nothing
+                   , errorStdOut = ""
+                   , errorStdErr = ""
                    }
 
 -- | The method called does not exist (from the JSON-RPC spec)
@@ -227,6 +241,8 @@ methodNotFound meth =
                    , message   = "Method not found"
                    , errorData = Just (JSON.toJSON meth)
                    , errorID   = Nothing
+                   , errorStdOut = ""
+                   , errorStdErr = ""
                    }
 
 -- | The request issued is not valid (from the JSON-RPC spec)
@@ -236,6 +252,8 @@ invalidRequest =
                    , message   = "Invalid request"
                    , errorData = Nothing
                    , errorID   = Nothing
+                   , errorStdOut = ""
+                   , errorStdErr = ""
                    }
 
 -- | Some unspecified bad thing happened in the server (from the JSON-RPC spec)
@@ -245,6 +263,8 @@ internalError =
                    , message   = "Internal error"
                    , errorData = Nothing
                    , errorID   = Nothing
+                   , errorStdOut = ""
+                   , errorStdErr = ""
                    }
 
 -- | Construct a 'JSONRPCException' from an error code, error text, and perhaps
@@ -256,6 +276,8 @@ makeJSONRPCException c m d =
     , message = m
     , errorData = JSON.toJSON <$> d
     , errorID = Nothing
+    , errorStdOut = ""
+    , errorStdErr = ""
     }
 
 {-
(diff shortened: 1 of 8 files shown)


argo/src/Argo.hs
```diff
@@ -339,6 +361,83 @@ instance JSON.ToJSON Request where
...
       "params"  .= view requestParams req
 
+data Response a
+  = Response { _responseID :: RequestID
+             , _responseAnswer :: a
+             , _responseStdOut :: String
+             , _responseStdErr :: String
+             } deriving (Eq, Ord, Show)
+
+responseID :: Simple Lens (Response a) RequestID
```
david-christiansen, line 371: `Simple Lens` isn't compatible with new GHCs. Can this be `Lens'` instead, which has the same meaning?

argo/src/Argo.hs
```diff
@@ -339,6 +361,83 @@ instance JSON.ToJSON Request where
...
+  ( \m reqID logger st ->
+      tryOutErr hCapture reqID (runMethod m logger st)
+  , \m logger st -> fst <$>
+      tryOutErr
+        (\hs a -> ("",) <$> hSilence hs a)
+        IDNull
+        (runMethod m logger st) )
+  where
+    tryOutErr ::
+      (forall a. [Handle] -> IO a -> IO (String, a)) ->
```
david-christiansen, line 419: Using `a` for two different type variables here confuses me.

argo/src/Argo.hs
```diff
@@ -339,6 +361,83 @@ instance JSON.ToJSON Request where
...
+      RequestID ->
+      IO (s, a) ->
+      IO (s, Response a)
+    tryOutErr capturer reqID action =
+      do (err, (out, res)) <-
+           capturer [stderr] . capturer [stdout] $
+           try @SomeException (try @JSONRPCException action)
+         tryPutOutErr out err
+         either
```
david-christiansen, line 428: I suspect this might be easier to follow as a `case` expression, but I don't feel strongly about it.

argo/src/Argo.hs
```diff
@@ -184,6 +186,10 @@ data JSONRPCException =
     , errorID :: Maybe RequestID
       -- ^ The request ID, if one is available. @Nothing@ omits
       -- it, because JSON-RPC defines a meaning for null.
+    , errorStdOut :: String
```
david-christiansen, line 189: I wonder if this should be `Maybe Text`.

My thought process is that `Text` is consistent with our other string types, and a quick `T.pack` on the stdout seems unlikely to make things appreciably slower (especially because we would need to traverse the whole thing anyway to send it over the wire). `Maybe` lets us distinguish between things that did not do stdio (e.g. by returning it as a JSON null). OTOH, it would never be `Just ""` (would it?), so this might be silly.

python/argo/connection.py
```diff
@@ -90,7 +88,7 @@ def pid(self) -> Optional[int]:
 
     def running(self) -> bool:
         """Check whether the process is still running."""
-        if self.proc is None or self.proc.poll() is None:
+        if self.proc is not None and self.proc.poll() is None:
```
david-christiansen, line 91: :-)

python/argo/interaction.py
```diff
@@ -6,21 +6,33 @@
...
-    message : str
-    data : Dict[str, Any]
-    code : int
 
+class ArgoException(Exception):
+    message: str
+    data: Dict[str, Any]
+    code: int
+    stdout: str
+    stderr: str
```
david-christiansen, line 20: I think str is the right thing here, but I want to be sure. What do we do if something dumps invalid UTF-8 (e.g. some literal bytes from some LLVM files)? Will the Haskell library behind the scenes catch it?

python/argo/interaction.py
```diff
@@ -81,21 +94,36 @@ def result(self) -> Any:
         pass
 
     @abstractmethod
-    def process_result(self, result : Any) -> Any:
+    def stdout(self) -> str:
```
david-christiansen, line 97: I think the convention for abstract methods is to have `...` as the body, rather than `pass`. I think I put the `pass`es in earlier, but we might as well fix it now.

python/argo/interaction.py
```diff
@@ -106,31 +134,46 @@ class Command(Interaction):
     corresponding command's appropriate representation.
     """
 
-    def _result_and_state(self) -> Tuple[Any, Any]:
+    def _result_and_state_and_out_err(self) -> Tuple[Any, Any, str, str]:
```
david-christiansen, line 137: I think that once a three-tuple is exceeded, it might be a good idea to make a little record type to hold it. The `dataclass` decorator makes it not too verbose in Python.

python/argo/interaction.py
```diff
@@ -106,31 +134,46 @@ class Command(Interaction):
...
-        return self._result_and_state()[1]
+        return self._result_and_state_and_out_err()[1]
 
     def result(self) -> Any:
         """Return the result of the command."""
-        return self.process_result(self._result_and_state()[0])
+        return self.process_result(self._result_and_state_and_out_err()[0])
+
+    def stdout(self) -> str:
+        """Return the stdout printed during the execution of the command."""
```
david-christiansen, line 166: Isn't it the _contents_  of `stdout` being returned, rather than the file handle itself? (pardon my pedantry)

python/saw/__init__.py
```diff
@@ -270,25 +321,28 @@ def llvm_verify(module: LLVMModule,
                                       assumptions=lemmas,
                                       contract=contract,
                                       exception=err)
-            abort_proof = True
+    # If something else went wrong...
+    except Exception as err:
+        __global_success = False
+        for view in __designated_views:
```
david-christiansen, line 327: What is it that makes the views "designated"? The name `designated_connection` uses it in the sense of "canonical", but here it seems to be those that the user selected. What about something like `__active_views` instead?

argo/src/Argo.hs
```diff
@@ -198,8 +204,12 @@ instance JSON.ToJSON JSONRPCException where
           Just theID -> JSON.toJSON theID
       , "error" .=
         (JSON.object $ ["code" .= JSON.Number (fromInteger (errorCode exn))
-                       , "message" .= JSON.String (message exn)] ++
-                       (("data" .=) <$> maybeToList (errorData exn)))
+                      , "message" .= JSON.String (message exn)
+                      , "data" .= (JSON.object $
+                        ([ "stdout" .= errorStdOut exn
+                         , "stderr" .= errorStdErr exn ] ++
+                         (("data" .=) <$> maybeToList (errorData exn))))
```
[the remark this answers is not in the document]
david-christiansen: Hmm, the difficulty coming up with a fitting name strikes me as a bit of a protocol wart. Perhaps we should just say that apps always return objects, and then merge in these two extra fields, clobbering any that the app made with those names?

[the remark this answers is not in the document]
kwf: I have a slight preference for nesting the field, since it means the application cannot possibly interfere with the field names. However, perhaps we could just name the two extra fields in a sufficiently obfuscated way that it would take intentional subversion to clash with them: for example, `__argo_method_stdout__` or some such. Do you think we need to worry that the client might clash with field names and create difficult-to-debug behavior?

argo/src/Argo.hs
kwf: I would expect so! I'll change it.

python/argo/connection.py
kwf: 😉

python/argo/interaction.py
kwf: Haskell `String`s are guaranteed to be valid UTF-8, aren't they? I would have to test this to be sure, but *something* ought to break and throw an exception—which would then be caught, wrapped, and returned to the client.

argo/src/Argo.hs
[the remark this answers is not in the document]
david-christiansen: I don't think obfuscating the names is that important. If we document that we'll clobber any fields with those names that the app developer inserted, then I don't think they'll be too surprised.

Putting myself in the shoes of an app developer who has seen those two fields in all of my tests, I would not be frustrated by them overwriting my own that I later inserted. We could also throw an exception if the developer included them that made it clear what had happened.

argo/src/Argo.hs
david-christiansen: Thanks!

python/argo/interaction.py
david-christiansen: Haskell Strings are not UTF-8, because they represent code points rather than bytes. But I don't think we can have a list of Haskell Chars that spits out invalid UTF-8 for Python to slurp in.

david-christiansen: Nothing says that stdio has characters in it on Unix, though - it's just bags of bytes, and UTF-8 is a convention. But the Haskell library that you're using to capture stdio takes care of this, right?

argo/src/Argo.hs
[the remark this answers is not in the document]
david-christiansen: I ended up resolving the issue thusly: `result` has 4 fields, `state`, `answer`, `stdout`, and `stderr`. This was made possible by making state tracking non-optional in #82.

david-christiansen:
I'm leaving a few comments here so I can remember to return to them tomorrow, but the important parts are now merged in and ready to go.

PR #82, opened 2020-07-01
Redo state management

david-christiansen:
States are now represented by opaque tokens (UUIDs) that are mapped behind the scenes to a cache of application states as a mix of actual reified saved states and recipes.

As part of getting all this to work, the submodules that are in common with saw-script were set to just use those in saw-script.

There is not yet a way to limit the size of the application state cache, and SAW actions other than Cryptol do not yet use the FS cache. I wanted to get Cryptol changes upstream as quickly as possible to avoid branches diverging too much.


python/argo/connection.py
```diff
@@ -142,6 +142,7 @@ def _buffer_replies(self) -> None:
         try:
             arrived = self.sock.recv(4096)
             while arrived != b'':
+                print(arrived)
                 self.buf.extend(arrived)
```
atomb, line 146: Is this a remnant of debugging?

argo/src/Argo/ServerState.hs
```diff
@@ -0,0 +1,161 @@
...
+recipeFile ::
+  (Eq recipeStep, Hashable recipeStep) =>
+  ServerState recipeStep s ->
+  [recipeStep] ->
+  FilePath -> IO ByteString
+recipeFile server recipe filename =
+  do hashes <- readIORef $ view stateFiles server
+     theHash <-
+       case view (at (recipe, filename)) hashes of
+         Nothing -> error "TODO exn"
```
atomb, line 131: Should this be a more useful error?

argo/src/Argo/ServerState.hs
```diff
@@ -0,0 +1,161 @@
...
+         Just h -> pure h
+     fileContents <- readIORef $ view stateFileContents server
+     case view (at theHash) fileContents of
+       Nothing -> error "TODO not found"
```
atomb, line 135: Same as the previous call to `error`.

argo/src/Argo.hs
```diff
@@ -83,13 +93,14 @@ jsonRPCVersion = "2.0"
 -- each of which is a function from the JSON value representing its parameters
 -- to a JSON value representing its response.
 newtype Method state result
-  = Method (ReaderT (Text -> IO ()) (StateT state IO) result)
+  = Method (ReaderT (Text -> IO (), FilePath -> IO B.ByteString) (StateT state IO) result)
```
kwf, line 96: Would it make sense to turn this tuple into a record with named fields, like:

``` haskell
data Config { logText :: Text -> IO (), readFile :: FilePath -> IO B.ByteString }
```

It seems like this could reduce potential confusion of the purpose of those things at use sites.

kwf: Although if the public interface in terms of `getDebugLogger` and `getFileReader` doesn't expose the `ReaderT` directly, this doesn't matter all that much.

argo/src/Argo.hs
```diff
@@ -297,7 +310,7 @@ data Request =
           -- distinct, we have both a null constructor and wrap it in a Maybe.
           -- When the request ID is not present, the request is a notification
           -- and the field is Nothing.
-          , _requestParams :: !JSON.Value
+          , _requestParams :: !JSON.Object
```
kwf, line 313: 👍

argo/src/Argo.hs
```diff
@@ -351,28 +376,70 @@ handleRequest logger respond app req =
...
-      do (_, answer) <- runMethod (m params) logger =<< readMVar theState
+      do stateID <- getStateID req
+         answer <- withMVar theState $
+           \contents ->
+             stateRecipe contents stateID >>=
+             \case
+               Nothing -> throwIO $ unknownStateID stateID
+               Just r ->
+                 do theAppState <- getAppState (rerunStep reqID app logger) contents r
+                    let fileReader = B.readFile -- No caching of this state
```
kwf, line 412: I'm not sure I understand what's happening here. Why do we cache `Command` state but not `Query` state?

argo/src/Argo/ServerState.hs
```diff
@@ -0,0 +1,180 @@
...
+newStateID :: IO StateID
+newStateID = StateID . Just <$> UUID.nextRandom
+
+instance JSON.ToJSON StateID where
+  toJSON (StateID (Just uuid)) = JSON.String (UUID.toText uuid)
+  toJSON (StateID Nothing) = JSON.Null
+
+instance JSON.FromJSON StateID where
+  parseJSON JSON.Null = pure $ StateID Nothing
+  -- Note: this is a backwards compatibility hack that should be removed soon
```
kwf, line 36: Should we emit a warning if this is used? Should we just break compatibility now, given that we currently are the only users?

argo/src/Argo/ServerState.hs
```diff
@@ -0,0 +1,180 @@
...
+  (Eq recipeStep, Hashable recipeStep) =>
+  (s -> recipeStep -> (FilePath -> IO ByteString) -> IO s) ->
+  ServerState recipeStep s ->
+  [recipeStep] ->
+  IO s
+runRecipe _ server [] = pure $ view initialAppState server
+runRecipe runStep server recipe@(step : steps) =
+  do s <- getAppState runStep server steps
+     let fileCache = recipeFile server recipe
+     runStep s step fileCache
```
kwf, line 180: It took me a second to parse out the mutual recursion between `getAppState` and `runRecipe`. Is there a way to phrase this that isn't mutually recursive? If not, it might be good to comment describing how they interact (I always find mutual recursion takes me a hot sec to load up in my head!).

argo/src/Argo/ServerState.hs
```diff
@@ -0,0 +1,180 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
```
kwf, line 1: This file seems the most critical to document for future maintainers, since it's where the handling of caching and mutable state happens, and that's the trickiest to reason about. I think it would be useful for future-you-and-me-and-others to add some explanatory comments for the fields of `ServerState` and top-level docs for the various functions in this module.

argo/src/Argo.hs
david-christiansen: I see the tuple as an internal implementation detail, rather than an interface. If it gets 3 things, then I'll think a record is in order :-)

argo/src/Argo/ServerState.hs
david-christiansen: We are not the only users - I know of at least one other. I think we should break the backwards compat soon, though.

argo/src/Argo/ServerState.hs
david-christiansen: We could do it by inlining them. Basically, one of them checks whether the recipe's state is cached, and the other reconstructs it if not. 

I'll at least add some docs.

argo/src/Argo/ServerState.hs
david-christiansen: Indeed - I agree completely and will do so.

argo/src/Argo.hs
david-christiansen: I believe that the comment in the newly-pushed version answers this.

Because a query by definition does not modify the state, we do not create a new server-side state, nor do we return a new state ID.

Issue #15, opened 2019-07-22
Add JVM support to the SAW API


david-christiansen:
The LLVM support should serve as a model for how to do this.

atomb:
The skeleton of the code exists, but two things remain:

* Some of the "language-independent" `SetupValue` stuff the LLVM interface uses hasn't been implemented for Java yet.
* The Java class loader doesn't use the system-wide, caching file loader yet.

PR #83, opened 2020-07-17
LLVM updates and the beginning of Java support

atomb:
This PR makes LLVM support more fully-featured (including support for x86 leaf function) and begins refactoring toward Java support. It's not ready to merge yet, but this PR helps review what remains before merging.


python/saw/exceptions.py
```diff
@@ -82,6 +84,8 @@ class CryptolError(SAWException): pass
     10050: NotAnLLVMMethodSpecIR,
     10060: NotASimpset,
     10070: NotATerm,
+    10080: NotAJVMClass,
```
david-christiansen, line 87: These should be added to `docs/Errors.rst` as well. Perhaps there's a good way to have one data structure that generates both the docs and these tables.

saw-remote-api/src/SAWServer.hs
```diff
@@ -59,27 +60,44 @@ import SAWServer.Exceptions
...
 data SAWTask
   = ProofScriptTask
-  | LLVMCrucibleSetup ServerName [SetupStep]
+  | LLVMCrucibleSetup ServerName [SetupStep LLVM.Type]
+  | JVMSetup ServerName [SetupStep JavaType]
 
 instance Show SAWTask where
   show ProofScriptTask = "ProofScript"
   show (LLVMCrucibleSetup n steps) = "(LLVMCrucibleSetup" ++ show n ++ " " ++ show steps ++ ")"
+  show (JVMSetup n steps) = "(JVMCrucibleSetup" ++ show n ++ " " ++ show steps ++ ")"
```
david-christiansen, line 73: The constructor doesn't include `Crucible`, but the string representation does. These should agree, right?

saw-remote-api/src/SAWServer/Data/Contract.hs
```diff
@@ -0,0 +1,87 @@
...
+    , allocatedAlignment  :: Maybe Int
+    }
+
+data PointsTo cryptolExpr =
+  PointsTo
+    { pointer  :: CrucibleSetupVal cryptolExpr
+    , pointsTo :: CrucibleSetupVal cryptolExpr
+    } deriving (Functor, Foldable, Traversable)
+
+instance FromJSON cryptolExpr => FromJSON (PointsTo cryptolExpr) where
```
david-christiansen, line 54: There's an implicit description of a data format in the protocol here. This should be added to `docs/Protocol.rst` eventually, if this design seems mostly ready to go. We haven't written the SAW part up yet due to flux, but it seems to be settling down at least a bit.

saw-remote-api/src/SAWServer/Data/SetupValue.hs
```diff
@@ -0,0 +1,51 @@
...
+      _ -> empty
+
+instance FromJSON cryptolExpr => FromJSON (CrucibleSetupVal cryptolExpr) where
+  parseJSON v =
+    nullValue v <|> fromObject v
+
+    where
+      nullValue = withText "setup value text" $
+        \case
+          "null" -> pure NullValue
```
david-christiansen, line 40: Just checking - do we want a protocol that uses the constant string `"null"` in a wire format with a constant named `null`? This seems like it might be a source of annoying bugs and misunderstandings. I don't think a JSON `null` makes sense here, but perhaps it should be some other string? And is `setup value text` what we want to see in parse errors?

saw-remote-api/src/SAWServer/JVMCrucibleSetup.hs
```diff
@@ -0,0 +1,197 @@
...
+data ServerSetupVal = Val (SetupValue CJ.JVM)
+
+-- TODO: this is an extra layer of indirection that could be collapsed, but is easy to implement for now.
+compileJVMContract ::
+  (FilePath -> IO ByteString) ->
+  BuiltinContext ->
+  CryptolEnv ->
+  Contract JavaType (P.Expr P.PName) ->
+  JVMSetupM ()
+compileJVMContract fileReader bic cenv c = interpretJVMSetup fileReader bic cenv (reverse steps)
```
david-christiansen, line 75: Why reverse the steps here, and then again later?

saw-remote-api/src/SAWServer/JVMCrucibleSetup.hs
```diff
@@ -0,0 +1,197 @@
...
+      do res <- liftIO $ getTypedTermOfCExp fileReader (biSharedContext bic) cenv expr
+         -- TODO: add warnings (snd res)
+         case fst res of
+           Right (t, _) -> return t
+           Left err -> error $ "Cryptol error: " ++ show err -- TODO: report properly
+
+    resolve env name =
+       case Map.lookup name env of
+         Just v -> return v
+         Nothing -> error "Server value not found - impossible!" -- rule out elsewhere
```
david-christiansen, line 177: Argo is now using the standard Galois Panic bit - perhaps this should call that?

saw-remote-api/src/SAWServer/JVMCrucibleSetup.hs
```diff
@@ -0,0 +1,197 @@
...
+      \case
+        Val x -> return x -- TODO add cases for the server values that
+                          -- are not coming from the setup monad
+                          -- (e.g. surrounding context)
+    getSetupVal (_, cenv) (CryptolExpr expr) = JVMSetupM $
+      do res <- liftIO $ getTypedTermOfCExp fileReader (biSharedContext bic) cenv expr
+         -- TODO: add warnings (snd res)
+         case fst res of
+           Right (t, _) -> return (MS.SetupTerm t)
+           Left err -> error $ "Cryptol error: " ++ show err -- TODO: report properly
```
david-christiansen, line 160: Presumably "report properly" means to throw a JSON-RPC exception. This seems like something that could be a user mistake rather than a server developer's mistake.

saw-remote-api/src/SAWServer/JVMCrucibleSetup.hs
```diff
@@ -0,0 +1,197 @@
...
+jvmLoadClass :: JVMLoadClassParams -> Method SAWState OK
+jvmLoadClass (JVMLoadClassParams serverName cname) =
+  do tasks <- view sawTask <$> getState
+     case tasks of
+       (_:_) -> raise $ notAtTopLevel $ map fst tasks
+       [] ->
+         do bic <- view sawBIC <$> getState
+            c <- tl $ loadJavaClass bic cname
+            setServerVal serverName c
+            -- TODO: trackFile
```
david-christiansen, line 196: We should actually delete the trackFile stuff. It's obsoleted by #82.

saw-remote-api/src/SAWServer/JVMCrucibleSetup.hs
```diff
@@ -0,0 +1,197 @@
...
+    JVMLoadClassParams <$> o .: "name" <*> o .: "class"
+
+jvmLoadClass :: JVMLoadClassParams -> Method SAWState OK
+jvmLoadClass (JVMLoadClassParams serverName cname) =
+  do tasks <- view sawTask <$> getState
+     case tasks of
+       (_:_) -> raise $ notAtTopLevel $ map fst tasks
+       [] ->
+         do bic <- view sawBIC <$> getState
+            c <- tl $ loadJavaClass bic cname
```
david-christiansen, line 194: Since #82, `loadJavaClass` should be using the IO action returned from `getFileReader` so that the caching support will be transparent.

saw-remote-api/src/SAWServer/JVMVerify.hs
```diff
@@ -0,0 +1,65 @@
...
+import SAWServer
+import SAWServer.Data.JVMType
+import SAWServer.Exceptions
+import SAWServer.JVMCrucibleSetup
+import SAWServer.OK
+import SAWServer.ProofScript
+import SAWServer.TopLevel
+import SAWServer.VerifyCommon
+
+jvmVerify :: VerifyParams JavaType -> Method SAWState OK
```
david-christiansen, line 29: This function and `jvmAssume` are incredibly similar. Should we worry about them getting out of sync long-term? If so, does it make sense to have them call common helpers to set up and tear down the verification task?

saw-remote-api/src/SAWServer/VerifyCommon.hs
```diff
@@ -0,0 +1,91 @@
...
+    , verifyLemmas       :: [ServerName]
+    , verifyCheckSat     :: Bool
+    -- TODO: might want to be able to save contracts and refer to them by name?
+    , verifyContract     :: Contract ty Expression -- ServerName
+    -- TODO: might want to be able to save proof scripts and refer to them by name?
+    , verifyScript       :: ProofScript
+    , verifyLemmaName    :: ServerName
+    }
+
+data X86Alloc = X86Alloc String Integer
```
david-christiansen, line 31: Haddocks would be nice here

david-christiansen:
Also, what about at least a very low-level Python script to test this?

python/saw/exceptions.py
atomb: Ah, good. I'll add those.

saw-remote-api/src/SAWServer.hs
atomb: Oh, yes, they should.

saw-remote-api/src/SAWServer/Data/Contract.hs
atomb: Yeah, let's write that up soon. I've created #84 to track that.

saw-remote-api/src/SAWServer/Data/SetupValue.hs
atomb: This is a remnant from old code that's just showing up as added because it's indented more. I was actually wondering about it, too. It's not only potentially confusing text, but also somewhat inconsistent with the other setup values. If you don't know of any reason to leave it how it is, I'll change it to make it more consistent and less confusing.

saw-remote-api/src/SAWServer/JVMCrucibleSetup.hs
atomb: Oh, good catch. This is copied from the LLVM interface, so it probably needs to be changed there, too.

saw-remote-api/src/SAWServer/JVMCrucibleSetup.hs
atomb: Fixed.

saw-remote-api/src/SAWServer/JVMVerify.hs
atomb: Good point. I abstracted something out for both the JVM and LLVM code.

saw-remote-api/src/SAWServer/VerifyCommon.hs
atomb: Done!

saw-remote-api/src/SAWServer/JVMCrucibleSetup.hs
atomb: Yeah, it should, but I believe it depends on code that hasn't been refactored to use that mechanism yet. The Java verification commands aren't functional yet, in other ways, so I'm inclined to hold off on implementing that. I added a comment to #15 to track this.

saw-remote-api/src/SAWServer/JVMCrucibleSetup.hs
atomb: I think so, though it'll require some refactoring. I created #85 to track that.

saw-remote-api/src/SAWServer/JVMCrucibleSetup.hs
atomb: I think so, though it'll require some refactoring. I created #85 to track that.

saw-remote-api/src/SAWServer/JVMVerify.hs
david-christiansen: Great, thanks!

saw-remote-api/src/SAWServer/JVMCrucibleSetup.hs
david-christiansen: Sounds reasonable to me.

saw-remote-api/src/SAWServer/JVMCrucibleSetup.hs
david-christiansen: :cold_sweat:

docs/Errors.rst
```diff
@@ -64,6 +64,11 @@ Server value errors (``10000``–``10099``)
 - ``10050``: “The server value with name \____\_ is not an LLVM method
    specification”
    ``{ name: String }``
+- ``10080``: “The server value with name \____\_ is not an JVM class”
```
david-christiansen, line 67: I think this should be "a JVM class" rather than "an JVM class"

saw-remote-api/src/SAWServer/Data/SetupValue.hs
david-christiansen: The new version is much nicer, both as code and as interface.

david-christiansen:
With the exception of a couple of English errors, I think it's good to merge.

I worry a bit about not including even a trivial low-level API test, because that kind of thing is easier now than later, but I don't think it's the end of the world to wait.

docs/Errors.rst
atomb: Fixed.

PR #92, opened 2020-10-07
Move `cryptol-remote-api` to the Cryptol repo

atomb:
In conjunction with [a related PR](https://github.com/GaloisInc/cryptol/pull/927), this moves the `cryptol-remote-api` package from the `argo` repo to the `cryptol` repo.


tasty-script-exitcode/src/Test/Tasty/HUnit/ScriptExit.hs
```diff
@@ -82,7 +84,14 @@ withPython3venv ::
   IO a
 withPython3venv requirements todo =
   withSystemTempDirectory "virtenv" $ \venvDir ->
-  do let process = proc "python3" ["-m", "venv", venvDir]
+  do mpy3 <- findExecutable "python3"
```
david-christiansen, line 87: This bit of code is presumably here to allow the CI to have different names for the Python executable, and not fail if we guess wrong. Is that right? If so, I think a little explanatory comment to that effect would be nice.

tasty-script-exitcode/src/Test/Tasty/HUnit/ScriptExit.hs
```diff
@@ -109,8 +118,16 @@ withPython3venv requirements todo =
                    ":\nstdout: " <> pipStdout <> "\nstderr: " <> pipStderr
                  (ExitSuccess, _, _) ->
                    pure ()
-         in do traverse (\reqPath -> pip ["install", "-r", reqPath]) requirements
-               todo pip venvPython
+         in do (exitCode, stdout, stderr) <- readCreateProcessWithExitCode pipInstall ""
+               case exitCode of
+                 ExitFailure code ->
+                   assertFailure $
+                   "Failed to create install `pip` with code " <>
```
david-christiansen, line 125: I'm not sure what "create install" means here, and I would be quite confused with this test failure.

tasty-script-exitcode/tasty-script-exitcode.cabal
```diff
@@ -29,13 +29,13 @@ library
...
   other-extensions:    NamedFieldPuns
-  build-depends:       base >=4.12 && <4.14,
-                       process ^>=1.6.3.0,
-                       filepath ^>=1.4.2.1,
-                       directory ^>=1.3.3.0,
-                       containers ^>=0.6.0.1,
-                       tasty-hunit ^>= 0.10,
-                       tasty ^>= 1.2,
-                       temporary ^>= 1.3
+  build-depends:       base >=4.12,
```
david-christiansen, line 32: The purpose of the `^>=` is to add a PVP-based upper bound. This change asserts that we'll work with all the future versions of these packages - is that what we want to be saying?

david-christiansen (changes requested):
The deletions all make sense.

The changes to the test code for CI are good to have, but it feels a bit like they're in the "hack until it works" category rather than the "clean thing that works" category. I think that at least an explanatory comment or three could help us make sure we keep them working in the future.

tasty-script-exitcode/tasty-script-exitcode.cabal
atomb: I'm not sure that I agree that that's what it means. In my opinion it means that we don't know that we _don't_ work with future versions and don't want to rule them out until we do know that they'l cause problems.

tasty-script-exitcode/src/Test/Tasty/HUnit/ScriptExit.hs
atomb: Yep, exactly. I added a comment describing that.

tasty-script-exitcode/src/Test/Tasty/HUnit/ScriptExit.hs
atomb: Oops! That definitely doesn't make sense. :) Fixed!

tasty-script-exitcode/tasty-script-exitcode.cabal
david-christiansen: I was going off the descriptions here: https://cabal.readthedocs.io/en/3.4/cabal-package.html#build-information

But the Platonic form of meaning of the operators and their realization in practice is certainly different, and in practice, we'll have bounds imposed by Cryptol itself anyway.

tasty-script-exitcode/src/Test/Tasty/HUnit/ScriptExit.hs
david-christiansen: Great!

david-christiansen (approved):

PR #93, opened 2020-10-07
Move `saw-remote-api` to the SAW repo 

atomb:
In conjunction with a [related PR](https://github.com/GaloisInc/saw-script/pull/853), this moves the `saw-remote-api` package from the `argo` repo to the `saw-script` repo.


david-christiansen (commented):
Comments here are really just those on #92.

PR #98, opened 2020-10-09
Add hostname option

jared-w:
Set the PR into this branch because that's what saw-script is currently pulling in as a submodule.

### DIFF
--- a/argo/src/Argo/DefaultMain.hs
+++ b/argo/src/Argo/DefaultMain.hs
@@ -36,7 +36,7 @@ newtype Port = Port String deriving (Eq, Ord)
 data TransportOpt
   = StdIONetstring
   -- ^ NetStrings over standard IO
-  | SocketNetstring (Maybe Session) (Maybe Publicity) (Maybe Port)
+  | SocketNetstring (Maybe Session) (Maybe Publicity) (Maybe Port) (Maybe HostName)
   -- ^ NetStrings over some socket
 
 data Publicity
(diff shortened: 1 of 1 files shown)


argo/src/Argo/DefaultMain.hs
```diff
@@ -166,18 +173,20 @@ realMain theApp opts =
   case transportOpt opts of
     StdIONetstring ->
       serveStdIONS theApp
-    SocketNetstring sessionOpt publicityOpt portOpt ->
+    SocketNetstring sessionOpt publicityOpt portOpt hostNameOpt ->
       do sessionResult <- getOrLockSession sessionOpt publicityOpt portOpt
+         let hostname = case hostNameOpt of
```
david-christiansen, line 178: As a matter of style, I very weakly prefer `fromMaybe` for this pattern. But it's not a big deal.

david-christiansen (commented):
This looks good to me.

@atomb, do you think this should go onto the designated branch, or into `master`? I don't have a lot of insight into how you're working through the server code transition, or what would be most convenient there.

argo/src/Argo/DefaultMain.hs
```diff
@@ -77,6 +77,13 @@ transport =
        Opt.help "Create or look up a globally available session")
       <|> pure Nothing
 
+    hostname =
+      (fmap Just . Opt.strOption $
+        Opt.long "host"
+        <> Opt.metavar "HOST"
+        <> Opt.help "Make server available at the secified hostname")
```
kquick, line 84: specified

david-christiansen:
Now that #93 and #92 are merged, I do think this should target `master` instead.

jared-w:
I updated it to target master instead.

david-christiansen:
Thank you!

PR #99, opened 2020-10-09
echo server example

pnwamk:
Adds a simple echo server example usage of argo that has simple commands and queries and leverages the server's file reading capabilities.

Potential Remaining Tasks
- [ ] add some tests for the argo error handling related to invalid JSON-RPC requests? E.g., parse error (-32700), invalid request (-32600), internal error (-32603), and unknown state id (20). N.B., internal error could be tested by just making a query function that simply errors on the example server and calling that at the end of our test suite, but the others might need to circumvent (or get creative with) the python argo API (since the python argo lib tries to avoid making/allowing the other errors I believe)
- [ ] incorporate `cabal test file-echo-api` into our CI for this repo


david-christiansen (changes requested):
This all looks good, and I think it'll be a valuable example to have.

One request: we use ReST for all the rest of the docs, due to Sphinx and Python, including the README. Can the docs here be ReST instead of .md?

david-christiansen:
How about merging this and then converting the remaining tasks into issues?

david-christiansen (approved):

PR #101, opened 2020-10-14
Initial cut at HTTP (WIP)

david-christiansen:
Remaining to be done:
 - [x] Validate MIME-type and such in headers, in accord with JSON-RPC site
 - [x] Develop some non-Python tests, to help us distinguish Python binding bugs from server bugs
 - [x] Documentation

### DIFF
--- a/argo/src/Argo.hs
+++ b/argo/src/Argo.hs
@@ -699,11 +699,12 @@ serveHandlesNS hLog hIn hOut app =
 
 serveHTTP ::
   HasCallStack =>
-  App s  {- JSON-RPC app -} ->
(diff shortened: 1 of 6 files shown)


argo/src/Argo.hs
```diff
@@ -699,11 +699,12 @@ serveHandlesNS hLog hIn hOut app =
 
 serveHTTP ::
```
pnwamk, line 700: Would `serveHttp` better match the rest of the code?

david-christiansen: Done

pnwamk (approved):

Issue #31, opened 2019-11-14
Add unit tests exercising each error code


kwf:
We want to make sure the errors we expect are the ones we get, so it would be good to add some expected-to-fail unit tests.

pnwamk:
Commit 5dcb409170e0c9dc6df91aa754072da63849a400 has partially addressed this (via unit tests in the `file-echo-api` unit tests), however, the below listed error codes could still use some manual checks.

- [ ] [parse error](https://github.com/GaloisInc/argo/blob/5dcb409170e0c9dc6df91aa754072da63849a400/argo/src/Argo.hs#L237) (-32700)
- [ ] [invalid request](https://github.com/GaloisInc/argo/blob/5dcb409170e0c9dc6df91aa754072da63849a400/argo/src/Argo.hs#L259) (-32600)
- [ ] [internal error](https://github.com/GaloisInc/argo/blob/5dcb409170e0c9dc6df91aa754072da63849a400/argo/src/Argo.hs#L270) (-32603)
- [ ] [unknown state id](https://github.com/GaloisInc/argo/blob/5dcb409170e0c9dc6df91aa754072da63849a400/argo/src/Argo.hs#L282) (20)

N.B., some of these error codes may be better tested directly (i.e., from a simple Haskell unit test) if the relevant portion of the  python `argo` API is designed to only send correct, well-formed requests.

PR #102, opened 2020-10-15
Additional negative tests

pnwamk:
See https://github.com/GaloisInc/argo/issues/31

### DIFF
--- a/file-echo-api/README.rst
+++ b/file-echo-api/README.rst
@@ -20,6 +20,11 @@ Commands
   * Clears any loaded file from the server's memory.
   * No parameters
 
+* ``implode``
+  
+  * Causes the server to raise an internal error (for testing purposes).
+  * No parameters
+
 Queries
 -------------------------
 
--- a/file-echo-api/file-echo-api/Main.hs
+++ b/file-echo-api/file-echo-api/Main.hs
@@ -29,5 +29,6 @@ serverMethods :: [(Text, Argo.MethodType, JSON.Value -> Argo.Method FES.ServerSt
 serverMethods =
   [ ("load",           Argo.Command,   Argo.method FES.loadCmd)
   , ("clear",          Argo.Command,   Argo.method FES.clearCmd)
+  , ("implode",        Argo.Query,     Argo.method FES.implodeCmd)
   , ("show",           Argo.Query,     Argo.method FES.showCmd)
   ]
--- a/file-echo-api/test-scripts/file-echo-tests.py
+++ b/file-echo-api/test-scripts/file-echo-tests.py
@@ -5,6 +5,7 @@ import signal
 import subprocess
 import sys
 import time
+import json
 
 import argo.connection as argo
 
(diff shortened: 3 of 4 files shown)


file-echo-api/README.rst
```diff
@@ -20,6 +20,11 @@ Commands
   * Clears any loaded file from the server's memory.
   * No parameters
 
+* ``implode``
```
david-christiansen, line 23: Excellent name :)

file-echo-api/file-echo-api/Main.hs
```diff
@@ -29,5 +29,6 @@ serverMethods :: [(Text, Argo.MethodType, JSON.Value -> Argo.Method FES.ServerSt
 serverMethods =
   [ ("load",           Argo.Command,   Argo.method FES.loadCmd)
   , ("clear",          Argo.Command,   Argo.method FES.clearCmd)
+  , ("implode",        Argo.Command,   Argo.method FES.implodeCmd)
```
david-christiansen, line 32: Why is this a command, if the server never enters a new state through implosion? (yes, nit-picky)

file-echo-api/test-scripts/file-echo-tests.py
```diff
@@ -95,6 +96,35 @@ def run_tests(c):
...
+    uid = c.send_message("show", {"state": "12345678-9101-1121-3141-516171819202"})
+    actual = c.wait_for_reply_to(uid)
+    expected = {'error':{'data':{'stdout':'','data':'12345678-9101-1121-3141-516171819202','stderr':''},'code':20,'message':'Unknown state ID'},'jsonrpc':'2.0','id':uid}
+    assert(actual == expected)
+
+    # invalid request (missing JSON-RPC required fields)
+    invalid_request = {'jsonrpc': '2.0','method': 'bad request', 'id':1}
+    c.process.send_one_message(json.dumps(invalid_request))
+    actual = c.wait_for_reply_to(None)
+    expected = {'error':{'data':{'stdout':'','data':'Error in $: key \"params\" not found','stderr':''},'code':-32700,'message':'Parse error'},'jsonrpc':'2.0','id':None}
```
david-christiansen, line 116: I worry that we're tying this too closely to the precise text of the Aeson-generated error message. In this case, I'd instead assert that the code and message fields were what's expected.

david-christiansen (commented):
I'm a bit concerned about hard-coding some internal auto-generated error message texts. But perhaps that is fine, and we'll just update the tests as it goes. What do you think?

file-echo-api/test-scripts/file-echo-tests.py
pnwamk: I don't entirely disagree.

Though, in this case would the "precise text of the Aeson-generated error message" here be something a user would potentially encounter? If so... it _might_ be nice to explicitly track/acknowledge it in some way and know when we change it (for better or worse).

pnwamk: > I'm a bit concerned about hard-coding some internal auto-generated error message texts. But perhaps that is fine, and we'll just update the tests as it goes. What do you think?

I think I'm inclined to want _some_ testing of error message texts that users might see, but limit it to as few such tests as necessary to reasonably track that particular aspect of the human-computer interaction. I totally agree if we had many such tests then that would mean our testing was overly repetitive and brittle, which would just be a pain in the long run and not be helpful.

pnwamk: After a chat, we decided to leave the test as is, with the thought we may want to improve these error messages slightly in the future (remove the `$:` meaning "top-level parsing object" for example)

Issue #110, opened 2020-10-28
Argo Python client and BitVector function arguments


weaversa:
cryptol-remote-api is giving a type mismatch error when providing BitVector types as arguments to Cryptol functions. I think the problem is here:

```cryptoltypes.py:152
        elif isinstance(val, BitVector.BitVector):
            copy = val.deep_copy()
            copy.pad_from_left(copy.length() % 4)
            return {'expression': 'bits',
                    'encoding': 'base64',
                    'width': val.length(), # N.B. original length, not padded                                                                 
                    'data': copy.get_bitvector_in_hex()}
```
If the encoding is in base64, we don't want the data to be in hex. So, one of those two things likely needs to change.

PR #111, opened 2020-10-28
fix: allow BitVector param type to Cryptol fns

pnwamk:
Should fix https://github.com/GaloisInc/argo/issues/110 - I've got a cryptol-remote-api test locally that successfully uses it, but nothing in this repo ... =\

### DIFF
--- a/python/cryptol/cryptoltypes.py
+++ b/python/cryptol/cryptoltypes.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 from collections import OrderedDict
 from abc import ABCMeta, abstractmethod
 import base64
+from math import ceil
 import BitVector #type: ignore
 
 from typing import Any, Dict, Iterable, List, NoReturn, Optional, TypeVar, Union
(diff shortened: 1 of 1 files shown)


python/cryptol/cryptoltypes.py
```diff
@@ -155,7 +155,7 @@ def convert(self, val : Any) -> Any:
             return {'expression': 'bits',
                     'encoding': 'base64',
                     'width': val.length(), # N.B. original length, not padded
-                    'data': copy.get_bitvector_in_hex()}
+                    'data': base64.b64encode(bytes(int(val))).decode('ascii')}
```
david-christiansen, line 158: I'm assuming that you've tested that this works round-trip. The sequence of conversions looks a bit mysterious :-)

Do we still need to pad the copy from the left if we're sending it as base64?

david-christiansen (changes requested):

python/cryptol/cryptoltypes.py
pnwamk: I had it wrong. After some googling around, manual experiments, and a better test from the cryptol-remote-api package I think this is right after 90df637.

pnwamk:
@david-christiansen if you think we need more testing I'm happy to add some somewhere, but things are looking good from manual poking and the corresponding BitVector test in the cryptol-remote-api PR here: https://github.com/GaloisInc/cryptol/pull/943

david-christiansen (approved):
This leaves us better off than before, at least :-)

Issue #100, opened 2020-10-13
Set up Github Actions


pnwamk:
to run `cabal test file-echo-api` as the basic unit test for Argo.

Obviously related to the desire to remove the irrelevant Travis CI tests: https://github.com/GaloisInc/argo/issues/91

pnwamk:
Fixed by #117

PR #117, opened 2020-11-10
chore: switch to github actions

pnwamk:


pnwamk:
Fixes https://github.com/GaloisInc/argo/issues/100

david-christiansen:
That green check mark is a thing of purest beauty

Issue #96, opened 2020-10-08
Improve Sphinx documentation extraction


ChrisEPhifer:
As of right now, we're naively extracting documentation from Python docstrings. This could be improved significantly by putting in some more manual work: Careful consideration of what double-underscore methods to include, what order to present the documentation, etc.

Personally, I like the guiding principle of "tell a story" - Even documentation of a module's interface should have a natural flow that makes its intended use clear.

PR #122, opened 2020-11-18
Documentation Sections

### LINKED ISSUE #94: Split documentation by components
The current documentation for Argo needs to be organized a bit better. Currently, the only split is between the protocol itself and the Python bindings - this could definitely be more fine-grained, especially in terms of separating the Cryptol-/SAW-specific APIs.

ChrisEPhifer:
Without changing any of the current contents of the documentation (other than typo fixes here and there), I've split them into more sections/files that seemed logical in order to resolve #94. High-level summary:

* Python bindings split into three files/sections: General-purpose bindings, Cryptol API, SAW API
* Protocol definition/description split into three files/sections: Overview (describing transport, state, representations, etc), Cryptol-specific protocol, [currently nonexistent] SAW-specific protocol

I'm of course open to suggestions on further stratification, and I imagine some of the work here will overlap with what I'm thinking about in terms of #96.


atomb (approved):
Seems like a good refactoring to me!

PR #125, opened 2020-11-23
Documentation of SAW Bindings

### LINKED ISSUE #95: Document the SAW bindings
There isn't really much documentation for the SAW bindings right now; there should be.

ChrisEPhifer:
This PR adds documentation (`docs/SAW.rst`) for the commands available for interacting with SAW. This includes not only the high-level commands, but descriptions of the JSON objects representing specifications/contracts, proof scripts, Crucible `Setup` values, LLVM types, and JVM types.

I believe this covers everything available in the API, based on the contents of `saw-remote-api/saw-remote-api/Main.hs` - I noticed there were some additional setup commands defined among the various modules, but since these aren't exposed in `Main.hs` I assumed they weren't available to users.

I would appreciate both feedback on typos and on organization - I attempted to mirror the structure of `docs/Cryptol.rst`, where high-level commands are introduced first, followed by specifications and `Setup` values, and finally LLVM/JVM types. This is similar to the placement of Cryptol types/terms toward the end of the Cryptol API documentation.

This PR _does not_ include documentation for any of the Python bindings.

Note that this resolves #95.


ChrisEPhifer:
FYI: I have a saw-script PR open to fold in the most recent updates to Cryptol (namely my addition of docs for the remote API). Once that and this are merged to their respective `master` branches, I'll be able to finalize the move of SAW documentation to saw-script.

docs/SAW.rst
```diff
@@ -2,9 +2,419 @@
...
+  are represented by a JSON object containing three fields:
+
+.. _contract-vars:
+
+  - ``server name``: The name of the variable on the server.
+  - ``name``: The "display name" of the variable, used in debugging output.
+  - ``type``: The :ref:`LLVM<llvm-types>` or :ref:`JVM<jvm-types>` type of this variable.
+
+``pre conds``
+  A list of the specification's preconditions, as Cryptol expressions.
```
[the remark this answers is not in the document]
atomb: Related to that, is there any way to link to the format for Cryptol terms from the Cryptol API documentation?

docs/SAW.rst
```diff
@@ -2,9 +2,419 @@
...
+  An optional :ref:`Crucible Setup value<setup-values>` specifying the expected return value of the function being verified.
+
+.. _proof-scripts:
+
+Proof Scripts
+=============
+
+SAW allows one to direct a verification task using a proof script, which is simply a sequence of proof
+tactics to apply. Very commonly, the proof script provided in a verification task is simply an instruction
+to use an external SAT/SMT solver such as ABS, Yices, or Z3.
```
atomb, line 246: ABS -> ABC

docs/SAW.rst
```diff
@@ -2,9 +2,419 @@
...
+    returns ``True``. This tactic fails if that is not the case.
+
+.. _setup-values:
+
+Crucible Setup Values
+=====================
+
+Setup Values encompass all values that can occur during symbolic execution, including Cryptol terms,
+pointers, arrays, and structures. They are used extensively when writing the specifications provided to the
+``verify`` commads. Setup Values are represented as JSON objects containing a tag field, ``setup value``,
```
atomb, line 289: commads -> commands

docs/SAW.rst
```diff
@@ -2,9 +2,419 @@
...
+
+``Cryptol``
+  A Cryptol term. There is an additional field ``expression`` containing a Cryptol expression.
+
+``array value``
+  An array value. There is an additional field ``elements`` which is a list of :ref:`Crucible Setup values<setup-values>`
+  to populate the array with.
+
+``field lvalue``
+  A struct field to be assigned to. There are two additional fields:
```
atomb, line 307: Reading this makes me realize that `field lvalue` and `element lvalue` aren't _only_ for destinations. They should perhaps have a different name.

docs/SAW.rst
```diff
@@ -2,9 +2,419 @@
...
+  - ``short``: A JVM short integer.
+
+``array type``
+  A JVM array. There are two additional fields:
+
+  - ``size``: An integer giving the length of the array.
+  - ``element type``: An :ref:`JVM type<jvm-types>` describing the array elements.
+
+``class type``
+  A JVM class. There is an additional field ``class name`` which identifies the class.
```
atomb, line 420: We should specify how the class name is encoded (dots or slashes), though I'm not sure the underlying code is finished enough to have solidified that decision.

atomb (changes requested):
Thanks for doing this! I noticed a few small typos and opportunities for clarification we should probably address before merging. The one big thing is that the code underlying the JVM methods is incomplete and doesn't yet work. I think the documentation for them that you've written is correct (w.r.t. to their intended behavior), so it would be sad to leave it out, but perhaps we can just add a temporary note that it's aspirational?

docs/SAW.rst
```diff
@@ -2,9 +2,419 @@
...
+------------
+
+:Method name:
+  ``SAW/Crpytol/save term``
+:Parameters:
+  - ``name``: The name to bind the value of ``expression`` to on the server.
+  - ``expression``: The Cryptol expression to bind the value of ``name`` to on the server.
+
+JVM Verification
+================
```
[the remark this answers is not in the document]
ChrisEPhifer: I added the following disclaimer right below this heading:

"NOTE: The following represents an aspirational view of the JVM-specific protocol; currently,
the code underlying these methods is incomplete and does not work."

docs/SAW.rst
```diff
@@ -2,9 +2,419 @@
...
+Loading Modules
+---------------
+
+:Method name:
+  ``SAW/LLVM/load module``
+:Parameters:
+  - ``name``: The name to bind the loaded bitcode file to on the server.
+  - ``bitcode file``: The path to the bitcode file to load and bind to ``name`` on the server.
+
+Verifying (General)
```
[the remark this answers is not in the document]
ChrisEPhifer: Much better :)

docs/SAW.rst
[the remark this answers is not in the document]
ChrisEPhifer: Good catch, I meant to add these links in :)

docs/SAW.rst
ChrisEPhifer: I didn't do anything with this one just yet - Perhaps I don't have a broad enough view of what an lvalue is, and that's more the problem than the name? :)

docs/SAW.rst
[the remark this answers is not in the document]
atomb: Perfect.

docs/SAW.rst
atomb: I should look at the code again and decide whether a better description is warranted.

docs/SAW.rst
ChrisEPhifer: From some quick spelunking, I _think_ class names are currently encoded using dots. I can update these docs to reflect that, if you'd like.

docs/SAW.rst
ChrisEPhifer: Sounds good!

ChrisEPhifer: So my own digging in to this suggests that you're right; these can be used as more than the targets of assignments. This is mostly an educated guess, though, based on naming and my (admittedly limited) understanding of the internals of Crucible :)

docs/SAW.rst
atomb: Sure. That sounds great.

docs/SAW.rst
atomb: Okay, cool. Then perhaps just "A field of a struct" instead of "A struct field to be assigned to" would be more appropriate. I think we can leave `lvalue` as part of the name, since that would require changing some code, though we could do that if it became confusing.

atomb (approved):
Looks good. Thanks for writing this up!

PR #132, opened 2020-12-09
Make stderr logging optional

david-christiansen:
This turns off debug logging to stderr by default, and adds a command-line option to enable it.

To future-proof the interface, logging is enabled by specifying a destination (which must be stderr for now) rather than by a Boolean flag.

### DIFF
--- a/argo/src/Argo/DefaultMain.hs
+++ b/argo/src/Argo/DefaultMain.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE OverloadedStrings, ScopedTypeVariables #-}
+{-# LANGUAGE LambdaCase, OverloadedStrings, ScopedTypeVariables #-}
 module Argo.DefaultMain (defaultMain) where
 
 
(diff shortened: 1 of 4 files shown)


argo/src/Argo/DefaultMain.hs
```diff
@@ -29,10 +29,12 @@ defaultMain str app =
      realMain app opts
 
 data NetworkOptions =
-  NetworkOptions (Maybe Session) (Maybe HostName) (Maybe Port)
+  NetworkOptions (Maybe Session) (Maybe HostName) (Maybe Port) (Maybe LogOption)
```
pnwamk, line 32: Does it make sense to use a record heir instead of positional arguments? 4 seems like a lot of positional params to me (although they do each have cute names to their types, so I could be easily disuaded for now I suppose)

pnwamk (approved):
LGTM - might be nice to use a record where I commented, although I trust your judgment since the types seem fairly distinct there as well.

david-christiansen:
I think the record is a good idea. That type keeps growing!

PR #133, opened 2020-12-09
Initial version of inline application documentation.

david-christiansen:
This is a breaking change for client applications, but not for their clients.

The idea is to gently guide application implementors into adding documentation at various points. Right now, it renders to ReST, but it can be extended to render to other formats in the future. I hope that it can be used to generate what Sphinx is getting as input, and that it will ease the task of keeping docs and code in sync.

Outputting plain text is an easy extension, while Markdown might be a bit harder due to the inability to arbitrarily nest Markdown docs. A best-effort "markdown" would be easy as well, though.

### DIFF
--- a/file-echo-api/src/FileEchoServer.hs
+++ b/file-echo-api/src/FileEchoServer.hs
@@ -2,14 +2,16 @@
 module FileEchoServer ( module FileEchoServer ) where
 
 import qualified Argo as Argo
-import           Control.Monad.IO.Class ( liftIO )
+import Argo.Doc
+import Control.Exception ( throwIO )
+import Control.Monad.IO.Class ( liftIO )
 import qualified Data.Aeson as JSON
-import           Data.Aeson ( (.:), (.:?), (.=), (.!=) )
-import           Data.ByteString ( ByteString )
+import Data.Aeson ( (.:), (.:?), (.=), (.!=) )
+import Data.ByteString ( ByteString )
 import qualified Data.ByteString.Char8 as Char8
 import qualified Data.Text as T
 import qualified System.Directory as Dir
-import           Control.Exception ( throwIO )
+
 
 
 newtype FileContents = FileContents String
@@ -67,13 +69,16 @@ instance JSON.FromJSON LoadParams where
     JSON.withObject "params for \"load\"" $
     \o -> LoadParams <$> o .: "file path"
 
+instance DescribedParams LoadParams where
+  parameterFieldDescription = [("file path", Paragraph [Text "The file to read into memory."])]
+
 loadCmd :: LoadParams -> Argo.Method ServerState ()
 loadCmd (LoadParams file) =
   do exists <- liftIO $ Dir.doesFileExist file
      if exists
      then do getFileContents <- Argo.getFileReader
              contents <- liftIO $ getFileContents file
-             Argo.setState $ ServerState 
+             Argo.setState $ ServerState
               { loadedFile = Just file
               , fileContents = FileContents $ Char8.unpack contents
               }
@@ -89,6 +94,9 @@ instance JSON.FromJSON ClearParams where
   parseJSON =
     JSON.withObject "params for \"show\"" $
     \_ -> pure ClearParams
+
+instance DescribedParams ClearParams where
+  parameterFieldDescription = []
 
 clearCmd :: ClearParams -> Argo.Method ServerState ()
 clearCmd _ =
@@ -113,6 +121,12 @@ instance JSON.FromJSON ShowParams where
              end <- o   .:? "end"
              pure $ ShowParams start end
 
+instance DescribedParams ShowParams where
+  parameterFieldDescription = [ ("start", Paragraph [Text "Start index (inclusive). If not provided, the substring is from the beginning of the file."])
+                              , ("end", Paragraph [Text "End index (exclusive). If not provided, the remainder of the file is returned."])
+                              ]
+
+
 showCmd :: ShowParams -> Argo.Method ServerState JSON.Value
 showCmd (ShowParams start end) =
   do (FileContents contents) <-  fileContents <$> Argo.getState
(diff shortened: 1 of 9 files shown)


david-christiansen:
Example of use:
```
» cabal run exe:file-echo-api -- doc                                                                                                                                  [14:56:15]
Up to date
File echo server
================

A sample server that demonstrates filesystem caching.

Datatypes
---------

.. _Ignorable:
Ignorable data
~~~~~~~~~~~~~~
Data to be ignored can take one of three forms:


``true``
  The first ignorable value
  
  
``false``
  The second ignorable value
  
  
``null``
  The ultimate ignorable value, neither true nor false
  
  Nothing else may be ignored.



Methods
-------

load (command)
~~~~~~~~~~~~~~


``file path``
  The file to read into memory.
  
  Load a file from disk into memory.


clear (command)
~~~~~~~~~~~~~~~

No parameters

Forget the loaded file.


implode (query)
~~~~~~~~~~~~~~~

No parameters

Throw an error immediately.


show (query)
~~~~~~~~~~~~


``start``
  Start index (inclusive). If not provided, the substring is from the beginning of the file.
  
  
``end``
  End index (exclusive). If not provided, the remainder of the file is returned.
  
  Show a substring of the file.


ignore (query)
~~~~~~~~~~~~~~


``to be ignored``
  The value to be ignored goes here.
  
  Ignore an :ref:`ignorable value <Ignorable>`.




```

file-echo-api/src/FileEchoServer.hs
```diff
@@ -2,14 +2,16 @@
 module FileEchoServer ( module FileEchoServer ) where
 
 import qualified Argo as Argo
-import           Control.Monad.IO.Class ( liftIO )
+import Argo.Doc
```
pnwamk, line 5: `import Argo.Doc` => `import qualified Argo.Doc as Doc` or similar IMO would be better. (Since part of the purpose of this module is to illustrate how to use Argo, erring on the side of verbose/pedantic imports I think might help some quickly grok things).

file-echo-api/src/FileEchoServer.hs
```diff
@@ -132,5 +146,45 @@ instance JSON.FromJSON ImplodeParams where
...
+  parameterFieldDescription = []
+
+
 implodeCmd :: ClearParams -> Argo.Method ServerState ()
 implodeCmd _ = liftIO $ throwIO $ Argo.internalError
+
+----------------------------------------------------------------------
+
+data Ignorable =
+  ThisDatatype | ExistsTo | DemonstrateDocs
```
pnwamk, line 159: lol

pnwamk:
Could we add some tests for this to CI?

pnwamk:
This is sweet. LGTM modulo the two comments I made.

file-echo-api/src/FileEchoServer.hs
david-christiansen: Sounds good. I'll make that change!

david-christiansen:
Do you have thoughts on the higher-level aspects of this design, before I spend time on fleshing it out with tests and the like?

For instance, there's a few "interesting" things about it. First off, it relies on explicit type applications to disambiguate the documentation type class, which I think is a bit of a code stench. The type class stuff is perhaps too cute? Or is it OK? Also, it requires documentation to create methods, which I think is good, but that doesn't extend to datatypes (like `Ignorable` in the demo, or cryptol expressions and contracts in the real apps).

Also, there's no easy online access to the documentation. This could be added fairly reasonably if you think it would be useful. For instance, we could add a message to all apps called "documentation" that returns a list of all methods and their docs. This would be nice for "generic" bindings, but I don't know how useful that is in practice.
 
Penultimately, what about integrating this into the Sphinx build script? The disadvantage of that is that it becomes harder to build the docs, but the advantage is that much of them is made automatic.

Finally, it should probably also emit overall argo-level protocol docs about state management and the like (these can be mostly copy-pasted).

file-echo-api/src/FileEchoServer.hs
david-christiansen: Done.

file-echo-api/src/FileEchoServer.hs
david-christiansen: :-D

pnwamk:
> For instance, there's a few "interesting" things about it. First off, it relies on explicit type applications to disambiguate the documentation type class, which I think is a bit of a code stench. The type class stuff is perhaps too cute? Or is it OK? 

Personally I do prefer my Haskell to be less "interesting" and "cute" generally... especially if the dumber version has good enough(TM) guarantees/usability. How painful would the less cute version be? (And if we keep the cuteness, maybe add some comments noting why there are unused type parameters on the associated type classes?)

> Also, it requires documentation to create methods, which I think is good, but that doesn't extend to datatypes (like `Ignorable` in the demo, or cryptol expressions and contracts in the real apps).

More required documentation would be nice.

> Also, there's no easy online access to the documentation. This could be added fairly reasonably if you think it would be useful. For instance, we could add a message to all apps called "documentation" that returns a list of all methods and their docs. This would be nice for "generic" bindings, but I don't know how useful that is in practice.

By `online` here you mean "while the Argo server is running"? These docs are for the client app author, right? I guess there's situations where querying for them would be useful.... but yah you'd imagine they might be used while writing the client and then never again. Yah... I'm not sure how much value added that is... 

> Penultimately, what about integrating this into the Sphinx build script? The disadvantage of that is that it becomes harder to build the docs, but the advantage is that much of them is made automatic.

TBH I haven't touch the Sphinx build script or doc generation much yet - how much harder would it be? Automatic sounds nice... ¯\\\_(ツ)\_/¯

david-christiansen:
> Personally I do prefer my Haskell to be less "interesting" and "cute" generally... especially if the dumber version has good enough(TM) guarantees/usability. How painful would the less cute version be? (And if we keep the cuteness, maybe add some comments noting why there are unused type parameters on the associated type classes?)

The less cute version would use `Proxy`. Basically it'd be a bit noisier - we'd replace `datatype @Foo` with `datatype (Proxy @Foo)` or `datatype (Proxy :: Proxy Foo)`.

> More required documentation would be nice.

Do you have thoughts on how to achieve this? I'm experimenting with designs for something that creates docs and parsers and unparsers out of the same code, and it's a bit... tricky so far. But I hope to get there!

> By online here you mean "while the Argo server is running"? These docs are for the client app author, right? I guess there's situations where querying for them would be useful.... but yah you'd imagine they might be used while writing the client and then never again. Yah... I'm not sure how much value added that is...

The thought process here is that we could write some Python code that queried the server and, at runtime, created low-level bindings that would auto-update as the server changed over time. Probably fairly non-useful. I'll leave that off.

david-christiansen:
> TBH I haven't touch the Sphinx build script or doc generation much yet - how much harder would it be? Automatic sounds nice... ¯\_(ツ)_/¯

It would not actually be harder per se, but it would mean that the docs couldn't be build by people who can't build the Haskell. This might make it harder to deploy the docs to hosting services like readthedocs.

pnwamk:
> The less cute version would use `Proxy`. Basically it'd be a bit noisier - we'd replace `datatype @Foo` with `datatype (Proxy @Foo)` or `datatype (Proxy :: Proxy Foo)`.

Ah - okay I'm not inclined to think that's an improvement, so maybe just adding some comments about seemingly unused type variables to the class declarations is good.

> > More required documentation would be nice.
> 
> Do you have thoughts on how to achieve this? I'm experimenting with designs for something that creates docs and parsers and unparsers out of the same code, and it's a bit... tricky so far. But I hope to get there!

Not at the moment - perhaps this is a good candidate for a Github issue we can revisit and collect thoughts on?

> It would not actually be harder per se, but it would mean that the docs couldn't be build by people who can't build the Haskell. This might make it harder to deploy the docs to hosting services like readthedocs.

...this doesn't sound that onerous to me, but I haven't had to worry about this much and probably don't have good intuition for the downstream costs for others.

atomb (approved):
I think this looks great. If we could get it so that a Sphinx run would automatically include the output, that would be awesome, but there's no need to hold things up based on that.

david-christiansen:
That should be a matter of updating the Sphinx makefile to spit out a ReST
file. But I'll see what I can fit in!

man. 4. jan. 2021 11.37 skrev Aaron Tomb <notifications@github.com>:

david-christiansen:
I think this is done enough to merge now, but I want to wait to get the clients updated before I actually push the button.

PR #138, opened 2020-12-15
Allow Argo apps to extend the command-line interface

david-christiansen:
Provide a more flexible "main" that allows Argo apps to enrich the
collection of allowed command-line arguments, and then construct the
application and initial state based on them.

This is part of https://github.com/GaloisInc/cryptol/issues/1009, and should probably not be merged until the Cryptol side has been implemented.

### DIFF
--- a/argo/src/Argo/DefaultMain.hs
+++ b/argo/src/Argo/DefaultMain.hs
@@ -25,20 +25,35 @@ import System.Exit (die)
 import Argo
 import Argo.Socket
 
+-- | The options selected by a user, including the server mode but
+-- excluding mode-specific initialization options (e.g. the HTTP port).
 data UserOptions stdIOOpts socketOpts httpOpts
   = StdIOOpts stdIOOpts
   | SocketOpts socketOpts
   | HttpOpts httpOpts
 
+-- | If all modes use the same user option, collapse them into a
+-- single structure.
 userOptions :: UserOptions a a a -> a
 userOptions (StdIOOpts opts) = opts
 userOptions (SocketOpts opts) = opts
 userOptions (HttpOpts opts) = opts
 
+-- | Run an Argo application using the default set of command-line
+-- arguments and server modes, augmented with additional command-line
+-- arguments for application-specific initialization.
+--
+-- For instance, this can be used to specify a file to load at server
+-- startup, or to customize server settings. The available settings
+-- can be different for each server mode, if desired.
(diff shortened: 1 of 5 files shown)


file-echo-api/file-echo-api/Main.hs
```diff
@@ -6,24 +6,37 @@ module Main ( main ) where
...
   "An RPC server for loading and printing files."
 
-mkInitState :: (FilePath -> IO ByteString) -> IO FES.ServerState 
-mkInitState = const $ FES.initialState
+mkInitState :: ServerOptions -> (FilePath -> IO ByteString) -> IO FES.ServerState
+mkInitState opts reader = FES.initialState (initialFile opts) reader
+
+newtype ServerOptions = ServerOptions { initialFile :: Maybe FilePath }
+
+parseServerOptions :: Opt.Parser ServerOptions
```
pnwamk, line 32: Might be nice to add a note here saying this function parses the _custom/additional_ options the server uses, and the handling of the standard options (e.g., http, stdio, socket) are still handled by the Argo backend.

pnwamk (approved):
Looks good to me. Would be nice to use/test the custom "load file on launch" feature in [file-echo-tests.py](https://github.com/GaloisInc/argo/blob/master/file-echo-api/test-scripts/file-echo-tests.py).

argo/src/Argo/DefaultMain.hs
```diff
@@ -21,14 +25,36 @@ import System.Exit (die)
...
+  | HttpOpts httpOpts
 
-defaultMain :: String -> App s -> IO ()
-defaultMain str app =
+userOptions :: UserOptions a a a -> a
+userOptions (StdIOOpts opts) = opts
+userOptions (SocketOpts opts) = opts
+userOptions (HttpOpts opts) = opts
+
+customMain ::
```
pnwamk, line 38: Now that there is both a custom and default main, it might be nice to have some simple Haddock to help newcommers and/or forgetful folks quickly grok the intended differences.

david-christiansen: Done! Good call.

file-echo-api/file-echo-api/Main.hs
david-christiansen: Done!

david-christiansen:
Test is added. Thanks!

PR #139, opened 2020-12-18
Add an option to avoid using `hCapture`

atomb:
This allows `argo` to work on a read-only file system. Or, at least, it should. If we discover any other file writes in the codebase, we can turn them off with the same option.

The one open question I have is whether the read-only option and the logger should be packaged together into some sort of more general record of options.

### DIFF
--- a/argo/src/Argo.hs
+++ b/argo/src/Argo.hs
@@ -67,7 +67,6 @@ import Control.Lens hiding ((.=))
 import qualified Data.Aeson as JSON
 import Data.Aeson ((.:), (.:!), (.=))
 import qualified Data.Aeson.Types as JSON (Parser, typeMismatch)
-import Data.Binary.Builder
 import qualified Data.ByteString as B
 import qualified Data.ByteString.Lazy as BS
 import Data.Foldable (for_)
@@ -77,12 +76,10 @@ import Data.Maybe (maybeToList)
 import Data.Scientific (Scientific)
 import Data.Text (Text)
 import qualified Data.Text as T
-import qualified Data.Text.IO as T
 import qualified Data.Text.Lazy as TL
 import GHC.Stack
 import Network.HTTP.Types.Method (StdMethod(..))
 import Network.HTTP.Types.Status
-import Network.Wai (strictRequestBody)
 import System.IO
 import System.IO.Silently
 import Text.Read (readMaybe)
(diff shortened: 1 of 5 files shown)


david-christiansen:
Shouldn't this be a standard part of the command line options parsed by customMain, rather than a top level Boolean flag passed to it?

david-christiansen:
And what about populating the stdout and stderr flags of return messages with null when this flag is enabled, just so clients can detect what's up?

atomb:
> Shouldn't this be a standard part of the command line options parsed by customMain, rather than a top level Boolean flag passed to it?

Perhaps so. My current thought was that `cryptol-eval-server` would have it unconditionally enabled, but I'm not wed to that approach.

atomb:
> And what about populating the stdout and stderr flags of return messages with null when this flag is enabled, just so clients can detect what's up?

Oh, yes, we should do that!

david-christiansen:
I think passing the default value in as an argument makes sense, but the
feature does perhaps become more useful for other servers if it can be
overridden.

pnwamk:
Is this something we can test easily with the echo server or similar? Might be nice to catch if down the road me make a tweak and `read-only` servers start getting chatty all of the sudden.

pnwamk (approved):
LGTM - would be nice to have a test in CI to check this works as expected going forward.

david-christiansen:
This is based on a quick skim while eating lunch - sorry!

One worry I have here is that the options record seems to contain two kinds
of thing: knobs turnable by the user (like avoiding hCapture) and internal
implementation details (like the file reader IO action). This means that
the default options record contains a file reader that's the wrong thing,
and that it can't be set by users. This, in my opinion, risks making the
code harder to understand.

What about making the method options record have two fields: implementation
internal stuff, and honest-to-goodness user preferences?

Make sense?

atomb:
Yeah, we should update the `file-echo-api` tests to cover this. And, speaking of that, its tests are failing because of the change to make `stdout` and `stderr` optional. I had missed that locally due to running `cabal test file-echo-api` without doing `cabal build file-echo-api` first. Given that, I'm going to hold off on merging this until I return in January.

atomb:
And, yeah, the file reader field was bugging me, too. I like the split between internal plumbing and real user preferences.

atomb:
I think I've addressed all of the issues except having a test that this really works on a read-only file system. Testing that will require some new infrastructure, so I'm going to create a separate issue for that, and we can address it in a separate PR.

Issue #136, opened 2020-12-10
Add support for cache eviction


atomb:

pnwamk:
See also https://github.com/GaloisInc/argo/issues/89 for some discussion on the topic.

pnwamk:
Here is some initial tinkering aimed at limiting memory usage which aims to _"give users the option to save a result state, but ensure incremental states don't sit around eating up resources"_ (as suggested by @weaversa in https://github.com/GaloisInc/argo/issues/89): https://github.com/GaloisInc/argo/tree/feature/bounded-memory-usage

## Approach Details

In that branch, Argo essentially aims to guarantee a bound on memory use by:
+ keeping only the _initial_ server state cached indefinitely,
+ keeping alive a (customizable) number of recently used states while evicting others from the cache when it fills (along with any no longer in-use file contents loaded from disk), and 
+ allows states to be explicitly pinned in the cache, which would keep them in memory until they are explicitly unpinned.

The main cost of such an approach is that arbitrary states no longer in cache cannot be "replayed" or "resurrected" automatically when a request mentioning them is received once they are evicted (such a state would either have to be manually pinned or rebuilt by the client through standard interactions).

## Other memory leaks =\

Unfortunately some preliminary testing suggests we still have some memory leaks even with the aforementioned changes when we should be able to maintain a relatively constant memory footprint. 

E.g., under a heavy load from a single client which should not have required continual increases in memory usage, the following two server setups still resulted in _significant_ memory usage while the Python client app used a reasonable and seemingly bounded amount of memory:
+ `argo.RemoteSocketProcess('localhost', 50005, ipv6=True)`
+ `argo.DynamicSocketProcess("cabal v2-exec file-echo-api --verbose=0 -- socket --port 50005")`

and these two setups had less significant but still persistently increasing memory usage under those same circumstances:
+ `argo.HttpProcess(url="http://localhost:8080/")`
+ `argo.StdIOProcess("cabal v2-exec file-echo-api --verbose=0 -- stdio")`

I haven't had a chance to hunt down these other memory leaks just yet. I suppose it's possible some of it is due to leaks in the server app itself (i.e., in the `file-echo-api` server implementation used for the tests), but the drastic difference between the two setups seems to suggest there is definitely some leaking that still needs addressed in Argo proper.

## Remaining Questions

[...]

PR #140, opened 2021-01-12
feat: bound memory usage by server

pnwamk:
This PR modifies Argo servers to guarantee a bound on memory usage.
This is accomplished by:
  1. keeping only the initial server state cached indefinitely,
  2. keeping alive a (customizable) number of recently used states
     while evicting others from the cache when it fills (along with
     any no longer in-use file contents loaded from disk), and
  3. allows states to be explicitly pinned in the cache, which would
     keep them in memory until they are explicitly unpinned.

The main cost of this change is that arbitrary states no longer
in cache cannot be "replayed" or "resurrected" automatically when
a request mentioning them is received once they are evicted from the cache
(such a state would either have to be manually pinned or rebuilt
by the client through standard interactions from a live state).

IF this change is a desirable step towards addressing https://github.com/GaloisInc/argo/issues/136, we may wish to add the following before merging:
+ [x] expose state cache pinning/unpinning/flushing capabilities to clients
+ [x] expose a method for querying which states are currently live (may or may not be useful)
+ [x] allow easy way to configure cache size when starting a server
+ [ ] document changes to `file-echo-api` (some methods were added to
        test the caching behavior)

### DIFF
--- a/argo/argo.cabal
+++ b/argo/argo.cabal
@@ -46,6 +46,7 @@ common deps
     scotty                      ^>= 0.12,
     silently                    ^>= 1.2,
     text                        ^>= 1.2.3,
+    time,
     unordered-containers        ^>= 0.2,
     uuid                        ^>= 1.3,
     wai                         ^>= 3.2.2,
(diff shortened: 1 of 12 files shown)


pnwamk:
@atomb I got this rebased and added a bunch of tests to sanity check the caching behavior and knobs behave as expected. I think it's basically ready to merge if you're happy with it.

atomb (approved):
I think this looks good!

Issue #146, opened 2021-01-28
Migrate Cryptol Python libraries to the `cryptol` repository


atomb:
Whenever we change `cryptol-remote-api` in the `cryptol` repository, the Python scripts in the `python` directory need updating. Ultimately, the `argo` repo should contain just the Python code for the core protocol of `argo`, and not the code for specific applications on top of it.

Issue #145, opened 2021-01-28
Migrate SAW Python libraries to the `saw-script` repository


atomb:
Whenever we change `saw-remote-api` in the `saw-script` repository, the Python scripts in the `python` directory need updating. Ultimately, the `argo` repo should contain just the Python code for the core protocol of `argo`, and not the code for specific applications on top of it.

PR #151, opened 2021-02-08
chore: remove cryptol- and saw-specific python code from argo repo

pnwamk:
Working towards:

+ [ ] migrating the cryptol specific python code to the `cryptol` repository
+ [ ] migrating the saw specific python code to the `saw-script` repository
+ [ ] publishing the `argo` client package on PyPi

See related issues https://github.com/GaloisInc/argo/issues/146 and https://github.com/GaloisInc/argo/issues/145.


pnwamk:
Seeing as this shouldn't hurt anyone even if we've mad a mistake... we're going to optimistically merge this PR now so it's easier to start experimenting with the downstream repos we're updating as a part of the overall effort.

PR #154, opened 2021-02-10
chore: add missing mypy annotations to argo-client package, fix tests

pnwamk:


python/tests/test_file_echo_api.py
```diff
@@ -306,45 +306,114 @@ def test_caching(self):
...
 class HttpTests(GenericFileEchoTests, unittest.TestCase):
+    # Connection to server
+    c = None
+    # process running the server
+    p = None
+
+    @classmethod
+    def setUpClass(self):
+        p = subprocess.Popen(
+            ["cabal", "run", "file-echo-api", "--verbose=0", "--", "http", "/", "--port", "8080"],
```
RyanGlScott, line 389: It's curious that other parts of this file use `cabal exec file-echo-api`, but this part uses `cabal run file-echo-api` instead. Is there a reason for doing something different here?

pnwamk: I've preferred `run` over `exec` after getting burned from the latter not taking into account whether things need rebuilt, etc, while the former does. Do you have a strong opinion and/or recommendation either way? I'd be happy to tweak them all to be whatever is "ideal" =)

pnwamk: I guess to directly answer your question, had I remembered I would have changed them all to use `run` to fit my personal preferences and avoid confusion down the line for people who run the tests and may have forgotten to explicitly rebuilt the Haskell executable.

RyanGlScott: I don't have a particularly strong opinion one way or the other, provided that you're happy. I'm mainly pointing this out since different tests in this file are now using `run` (e.g., `RemoteSocketProcessTests`) versus `exec` (e.g., `DynamicSocketProcessTests`) inconsistently.

If you do want to use `cabal run`, make sure you have a way to ignore the output from `cabal` itself so that it doesn't mistakenly get recognized as test output.

pnwamk: They now match  - if you think it would be better to not use `run` I'm all ears and happy to change things.

RyanGlScott: If `cabal run` works, then I'm quite happy to use that.

PR #155, opened 2021-02-22
feat: mutable file-echo-api with tests

pnwamk:
This PR adds support for servers with mutable state (like SAW's). This is accomplished as follows:

1. mkApp (i.e., the creation of an Argo server app) takes a `StateMutability` in the `AppOpts`, which indicates if initial states must be re-created for every new use (`MutableState`) or whether the same one can be reused for each new initial request (`PureState`).

2. `StateID` usage (for non-initial states) is now destructive, which will make preventing mutation in the underlying server from affecting other requests trivial. I.e., a when a command is made from a particular (non-initial) `StateID`, that `StateID` is destroyed and is no longer reachable, and a new `StateID` is associated with the resulting state from that request (regardless of whether or not the state actually changed). Down the road we can get more nuanced with this and keep states alive if they are in fact still valid and there is a desire to do so, but for now I think it keeps the implementation simple and safe and it meets all of the use cases we have encountered so far in practice.

3. The `Command`/`Query` distinction is gone -- for now all operations are `Command`s. Commands can touch state, and (as mentioned above) always destroy the `StateID` they are initiated from.

4. `Notification`s cannot touch individual states (i.e., they do not have corresponding `StateID`s), and are primarily used for server management (e.g., like when a connection is done performing calculations, a notification can "free" it's last `StateID`).

5. A `--max-occupancy` flag allows a server at launch to specify how many simultaneously live states may exist on the server at once (excluding the initial state). This, combined with proper "freeing" of states which are no longer needed (see (4)), should allow the server to not be overwhelmed beyond the specified max, while still allowing ongoing connections/operations as states are freed after use. Right now the state free-ing is manual (there is a way to free _all_ states as an escape hatch) -- later perhaps we want to make this smarter, add timeouts, etc, but I think this should suffice for now.

### DIFF
--- a/argo/src/Argo.hs
+++ b/argo/src/Argo.hs
@@ -32,9 +32,12 @@ module Argo
   -- * Defining methods
   , AppMethod
   , Method
+  , StatefulMethod
   , Command
+  , Query
   , Notification
   , command
+  , query
   , notification
   -- * Manipulating app state in methods
   , getState
@@ -115,7 +118,7 @@ jsonRPCVersion = "2.0"
 --------------------------------------------------------------------------------
 
 -- | Options affecting the how the server runs/supports the app.
-data AppOpts =
+newtype AppOpts =
   AppOpts
   {
     appStateMutability :: StateMutability
@@ -155,6 +158,12 @@ data CommandContext =
   , cmdCtxFileReader :: FilePath -> IO B.ByteString
   }
 
+data QueryContext =
+  QueryContext
+  { queryCtxMOptions :: MethodOptions
+  , queryCtxFileReader :: FilePath -> IO B.ByteString
+  }
+
 data NotificationContext =
   NotificationContext
   { ntfCtxMOptions :: MethodOptions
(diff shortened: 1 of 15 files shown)


argo/src/Argo.hs
```diff
@@ -108,6 +114,17 @@ jsonRPCVersion = "2.0"
 
 --------------------------------------------------------------------------------
 
+-- | Options affecting the how the server runs/supports the app.
+data AppOpts =
```
RyanGlScott, line 118: Make this a newtype, since this only has one field (for now, at least).

argo/src/Argo/DefaultMain.hs
```diff
@@ -110,22 +110,25 @@ data ProgramMode stdioOpts socketOpts httpOpts docOpts
 data GlobalOptions stdioOpts socketOpts httpOpts docOpts =
     GlobalOptions
     { methodOpts :: MethodOptions
+     -- ^ Max number of live concurrent states the server supports.
```
RyanGlScott, line 113: Isn't there more to `MethodOptions` that just the maximum number of live concurrent states? I feel like this comment ought to be a bit more general that just that.

pnwamk: Oops, yes that doesn't make sense. That's a copy+paste / code-reorg error on my part as I was experimenting with where to place things. Thanks!

PR #167, opened 2021-06-16
Bump upper version bounds on base to allow building with GHC 9.0

RyanGlScott:
Also add a GHC 9.0.1 configuration in CI.


robdockins (approved):
Looks like CI passes, I don't see any reason not to merge.

PR #174, opened 2021-07-15
feat: server can now be interrupted, cancelling all active requests

pnwamk:
This PR adds the ability to interrupt all active threads serving requests in the server.

Design notes:
+ For simplicity's sake, an interrupt is a server-wide event and cannot target specific requests. This seems reasonable as an initial constraint given that the `MVars` guarding the server state currently only allow one active request at a time. I.e., things are already fairly linear already so this shouldn't complicate any workflows.
+ For servers with *immutable state*, active requests which are cancelled have no observable effects since the updated state is only recorded upon completion of the request.
+ For servers with *mutable state* (or mutable _portions_ of state), effects (i.e., `IO`) from an interrupted thread may be observed since they cannot be rolled back in any meaningful way, however, the overall state (i.e., any immutable portions) will remain untouched for the same reason noted above.
+ `killThread` is currently used to kill all the active threads in the server. It is not clear how nicely this will interact with the Cryptol/SAW servers yet... in particular, killing a thread in this manner may leave lingering resources (e.g., external processes spawned by said thread) -- will have to follow up on this concern as we look to integrate the feature into the `cryptol-remote-api`

This should partially address https://github.com/GaloisInc/cryptol/issues/1147 i.e., we can interrupt long running commands, however, timeouts still need to be added separately.

### DIFF
--- a/argo/src/Argo/ServerState.hs
+++ b/argo/src/Argo/ServerState.hs
@@ -128,7 +128,7 @@ nextAppState server prevStateID newAppState = do
   modifyIORef' (serverStatePool server) $ HM.insert uuid newAppState
   return $ StateID uuid
 
--- | Desctroy an app state so it is no longer available for requests.
+-- | Destroy an app state so it is no longer available for requests.
 destroyAppState' ::
   ServerState s ->
   StateID ->
--- a/argo/src/Argo/ServerState.hs
+++ b/argo/src/Argo/ServerState.hs
@@ -18,6 +18,7 @@ module Argo.ServerState (
   StateID, initialStateID,
   ) where
 
+import Control.Concurrent
 import Numeric.Natural ( Natural )
 import Data.ByteString (ByteString)
 import qualified Data.ByteString as B
@@ -122,24 +123,35 @@ nextAppState ::
   appState {-^ The new application state -} ->
   IO StateID
 nextAppState server prevStateID newAppState = do
-  destroyAppState server prevStateID
+  destroyAppState' server prevStateID
   uuid <- UUID.nextRandom
   modifyIORef' (serverStatePool server) $ HM.insert uuid newAppState
   return $ StateID uuid
 
 -- | Desctroy an app state so it is no longer available for requests.
-destroyAppState ::
-  ServerState appState ->
+destroyAppState' ::
+  ServerState s ->
   StateID ->
   IO ()
-destroyAppState _server InitialStateID = pure ()
-destroyAppState server (StateID uuid) = modifyIORef' (serverStatePool server) $ HM.delete uuid
+destroyAppState' _server InitialStateID = pure ()
+destroyAppState' server (StateID uuid) = modifyIORef' (serverStatePool server) $ HM.delete uuid
+
+-- | Desctroy an app state so it is no longer available for requests.
+destroyAppState ::
+  MVar (ServerState s) ->
+  StateID ->
+  IO ()
+destroyAppState serverMVar sid =
+  withMVar serverMVar $ \server -> destroyAppState' server sid
+
 
 -- | Destroy all app states so they are no longer available for requests.
 destroyAllAppStates ::
-  ServerState appState ->
+  MVar (ServerState s) ->
   IO ()
-destroyAllAppStates server = writeIORef (serverStatePool server) $ HM.empty
+destroyAllAppStates serverMVar =
+  withMVar serverMVar $ \server ->
+    writeIORef (serverStatePool server) $ HM.empty
 
 -- | Retrieve the application state that corresponds to a given state ID.
 --
--- a/argo/src/Argo/ServerState.hs
+++ b/argo/src/Argo/ServerState.hs
@@ -114,9 +114,10 @@ statePoolCount :: ServerState appState -> IO Natural
 statePoolCount server =
   fromInteger . toInteger . HM.size <$> readIORef (serverStatePool server)
 
--- | Given a fresh application state and the StateID it was derived from,
--- return a fresh state ID that uniquely describes the new state and
--- destroy the previous state (if possible).
+-- | Given a fresh application state and the StateID it was derived from, return
+-- a fresh state ID that uniquely describes the new state and destroy the
+-- previous state (if possible). This ASSUMES the server state MVar is
+-- already locked elsewhere thus preventing inconsistent server states.
 nextAppState ::
   ServerState appState ->
   StateID {-^ State ID from which the new state originated (i.e., its parent) -} ->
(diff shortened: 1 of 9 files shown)


argo/src/Argo/ServerState.hs
```diff
@@ -122,24 +123,35 @@ nextAppState ::
   appState {-^ The new application state -} ->
   IO StateID
 nextAppState server prevStateID newAppState = do
-  destroyAppState server prevStateID
+  destroyAppState' server prevStateID
   uuid <- UUID.nextRandom
   modifyIORef' (serverStatePool server) $ HM.insert uuid newAppState
   return $ StateID uuid
 
+-- | Desctroy an app state so it is no longer available for requests.
```
RyanGlScott, line 131:
```suggestion
-- | Destroy an app state so it is no longer available for requests.
```

argo/src/Argo/ServerState.hs
```diff
@@ -122,24 +123,35 @@ nextAppState ::
...
+destroyAppState' ::
+  ServerState s ->
+  StateID ->
+  IO ()
+destroyAppState' _server InitialStateID = pure ()
+destroyAppState' server (StateID uuid) = modifyIORef' (serverStatePool server) $ HM.delete uuid
+
 -- | Desctroy an app state so it is no longer available for requests.
```
RyanGlScott, line 139:
```suggestion
-- | Destroy an app state so it is no longer available for requests.
```

argo/src/Argo/ServerState.hs
```diff
@@ -122,24 +123,35 @@ nextAppState ::
...
   IO StateID
 nextAppState server prevStateID newAppState = do
-  destroyAppState server prevStateID
+  destroyAppState' server prevStateID
   uuid <- UUID.nextRandom
   modifyIORef' (serverStatePool server) $ HM.insert uuid newAppState
   return $ StateID uuid
 
+-- | Desctroy an app state so it is no longer available for requests.
+destroyAppState' ::
```
RyanGlScott, line 132: If I understand correctly, the only difference between `destroyAppState` and `destroyAppState`' is that the latter will only destroy an app state if the ID is not initial? If so, it would be worth documenting this.

file-echo-api/src/FileEchoServer.hs
```diff
@@ -286,3 +289,75 @@ instance Doc.DescribedMethod DestroyAllStatesParams () where
 destroyAllStates :: DestroyAllStatesParams -> Argo.Notification ()
 destroyAllStates _ = Argo.destroyAllStates
 
+
+------------------------------------------------------------------------
+-- Sleep Query
+
+data SleepParams = SleepParams Int
```
RyanGlScott, line 296:
```suggestion
newtype SleepParams = SleepParams Int
```

file-echo-api/src/FileEchoServer.hs
```diff
@@ -286,3 +289,75 @@ instance Doc.DescribedMethod DestroyAllStatesParams () where
...
+
+
+interruptAllThreads :: InterruptAllThreadsParams -> Argo.Notification ()
+interruptAllThreads _ = Argo.interruptAllThreads
+
+
+------------------------------------------------------------------------
+-- SlowClear Command
+
+data SlowClear = SlowClear Int
```
RyanGlScott, line 342:
```suggestion
newtype SlowClear = SlowClear Int
```

file-echo-api/src/MutableFileEchoServer.hs
```diff
@@ -161,3 +164,78 @@ instance Doc.DescribedMethod DestroyStateParams () where
 
 destroyState :: DestroyStateParams -> Argo.Notification ()
 destroyState (DestroyStateParams stateID) = Argo.destroyState stateID
+
+
+------------------------------------------------------------------------
+-- Sleep Query
+
+data SleepParams = SleepParams Int
```
RyanGlScott, line 172:
```suggestion
newtype SleepParams = SleepParams Int
```

file-echo-api/src/MutableFileEchoServer.hs
```diff
@@ -161,3 +164,78 @@ instance Doc.DescribedMethod DestroyStateParams () where
...
+
+
+interruptAllThreads :: InterruptAllThreadsParams -> Argo.Notification ()
+interruptAllThreads _ = Argo.interruptAllThreads
+
+
+------------------------------------------------------------------------
+-- SlowClear Command
+
+data SlowClear = SlowClear Int
```
RyanGlScott, line 218:
```suggestion
newtype SlowClear = SlowClear Int
```

python/tests/test_file_echo_api.py
```diff
@@ -537,3 +538,102 @@ def test_occupancy_and_state_destruction(self):
...
+
+    @classmethod
+    def tearDownClass(self):
+        os.killpg(os.getpgid(self.p.pid), signal.SIGKILL)
+        super().tearDownClass()
+
+    # to be implemented by classes extending this one
+    def test_interrupts(self):
+        c1 = argo.ServerConnection(argo.HttpProcess('http://localhost:8083/'))
+        c2 = argo.ServerConnection(argo.HttpProcess('http://localhost:8083/'))
```
RyanGlScott, line 577: Can the `argo.HttpProcess('http://localhost:8083/')` bit be shared between the two lines above?

python/tests/test_mutable_file_echo_api.py
```diff
@@ -129,5 +137,104 @@ def test_subsequent_connection(self):
...
+
+    @classmethod
+    def tearDownClass(self):
+        os.killpg(os.getpgid(self.p.pid), signal.SIGKILL)
+        super().tearDownClass()
 
+    # to be implemented by classes extending this one
+    def test_interrupts(self):
+        c1 = argo.ServerConnection(argo.HttpProcess('http://localhost:8083/'))
+        c2 = argo.ServerConnection(argo.HttpProcess('http://localhost:8083/'))
```
RyanGlScott, line 176: Similarly, can the `HttpProcess` be shared here?

python/tests/test_file_echo_api.py
pnwamk: I don't think so - the HTTP requests are blocking, so to perform interrupts we need two independent connections.

python/tests/test_mutable_file_echo_api.py
pnwamk: Negative for same reason noted above.

python/tests/test_file_echo_api.py
RyanGlScott: To be clear, I'm not suggesting sharing the same connection, just the URL that each one connects to. I had something like this in mind:

```py
proc = argo.HttpProcess('http://localhost:8083/')
c1 = argo.ServerConnection(proc)
c2 = argo.ServerConnection(proc)
```

(Disclaimer: I haven't tested this.)

pnwamk: Unfortunately it's more than the URL contained there, i.e., `HttpProcess` has mutable state that can't be shared between `c1` and `c2` without breaking things. E.g., the `waiting_replies` member field.

argo/src/Argo/ServerState.hs
pnwamk: It's slightly more complicated... so I added a bunch of comments regarding these various scary effectful functions: https://github.com/GaloisInc/argo/pull/174/commits/4513fda6daaf5c075bfb1a16948cea4cec38e111

RyanGlScott (approved):
This looks quite thorough to me, especially given that there are test cases for the interruption feature. I like.

PR #179, opened 2021-07-22
feat: server and client logging options for RPC tx/rx

pnwamk:
This PR adds basic logging which can be toggled for both the argo client 
and server, s.t. each RPC transmission (TX) and reception (RX) is emitted
on a single line to the specified file/handle.

Tests ensuring this works as expected for each transport method available
have also been added to the python test suite.

The underlying goal here is to make it easier to debug server/client interactions
(especially in non-trivial deployment situations).

Any feedback/critique/etc on logging techniques used (e.g., is writing to a file
with `Text.appendFile` reasonable for file logging? etc) welcome!

### DIFF
--- a/argo/src/Argo.hs
+++ b/argo/src/Argo.hs
@@ -106,6 +106,7 @@ import Data.Set (Set)
 import qualified Data.Set as Set
 import Data.Text ( Text )
 import qualified Data.Text as T
+import qualified Data.Text.Encoding as T
 import qualified Data.Text.Lazy as TL
 import GHC.Stack
 import Numeric.Natural ( Natural )
@@ -165,6 +166,8 @@ data MethodOptions =
     -- ReadOnly prohibits creating or writing files.
     , optLogger :: Text -> IO ()
     -- ^ A function to use for logging debug messages.
+    , optLogFile :: Maybe FilePath
+    -- ^ Where things are being logged _if_ a log file was specified.
     , optMaxOccupancy :: Natural
     -- ^ How many simultaneous states can be live at once?
     }
@@ -174,6 +177,7 @@ defaultMethodOptions =
   MethodOptions
     { optFileSystemMode = ReadWrite
     , optLogger = const (return ())
+    , optLogFile = Nothing
     , optMaxOccupancy = 10
     }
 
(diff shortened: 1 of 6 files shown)


python/argo_client/connection.py
```diff
@@ -69,6 +72,22 @@ def get_one_reply(self) -> Optional[str]: pass
     @abstractmethod
     def send_one_message(self, the_message: str, *, expecting_response : bool = True) -> None: pass
 
+    def logging(self, on : bool, *, dest : TextIO = sys.stderr) -> None:
```
RyanGlScott, line 75: I'm not sure if this is intentional, but there is a widely used [`logging` module](https://docs.python.org/3/library/logging.html) in Python that this method could be confused for.

python/tests/test_file_echo_api.py
```diff
@@ -178,7 +204,7 @@ class RemoteSocketProcessTests(GenericFileEchoTests, unittest.TestCase):
     @classmethod
     def setUpClass(self):
         p = subprocess.Popen(
-            ["cabal", "run", "exe:file-echo-api", "--verbose=0", "--", "socket", "--port", "50005"],
+            ["cabal", "run", "exe:file-echo-api", "--verbose=0", "--", "socket", "--port", "50005", "--log", "remote-socket-server.log"],
```
RyanGlScott, line 207: Perhaps use `get_server_log_file()` here instead of duplicating the log file name. (Similarly for other test cases as well.)

argo/src/Argo.hs
```diff
@@ -1122,6 +1135,9 @@ serveHttp opts httpOpts app port = do
                    status $ if body == ""
                             then status204
                             else ok200
+                   liftIO $ if body == ""
+                            then logTx opts "[TX] 'NO CONTENT'"
```
atomb, line 1139: Will this duplicate the `[TX]` prefix?

atomb (approved):

python/argo_client/connection.py
pnwamk: Given that it's an instance method, so it will always be proceeded by `OBJECT.` for some `OBJECT`, I'm inclined to think it won't be confused with that module, however I'm more than happy to reconsider/change it if others disagree and prefer a different name.

RyanGlScott: I don't have a strong opinion on the matter—I just wanted wanted to the resemblance was noted. I'm quite fine with bucking convention :)

argo/src/Argo.hs
pnwamk: Ah, yes that would =) fix pushed

python/tests/test_file_echo_api.py
pnwamk: done

RyanGlScott (approved):

PR #183, opened 2021-08-26
feat: notifications delay new request handling while running

pnwamk:
Makes it so that the handling of a notification cannot be interrupted/interleaved with new commands/queries/notifications. This may be enough to fix https://github.com/GaloisInc/cryptol/issues/1261

Tweaking the default state eviction policy to remove older requests automatically instead of reject new ones if the server is full (thus removing the need for scripts to call `reset` when connecting) may also be helpful.

### DIFF
--- a/argo/src/Argo.hs
+++ b/argo/src/Argo.hs
@@ -98,12 +98,14 @@ import Data.Foldable (for_)
 import Data.IORef
 import Data.List.NonEmpty (NonEmpty(..))
 import Data.Map (Map)
-import qualified Data.Map as M
+import qualified Data.Map as Map
 import qualified Data.HashMap.Strict as HM
 import Data.Maybe (maybeToList)
 import Data.Scientific (Scientific)
 import Data.Set (Set)
 import qualified Data.Set as Set
+import Data.Sequence (Seq)
+import qualified Data.Sequence as Seq
 import Data.Text ( Text )
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
@@ -117,6 +119,10 @@ import System.IO.Silently ( hCapture, hSilence )
 import System.Environment (lookupEnv)
 import Text.Read (readMaybe)
 import Web.Scotty
+    ( scottyApp,
+      scottyOpts,
+      ScottyM)
+import qualified Web.Scotty as Scotty
     ( addroute,
       body,
       finish,
@@ -124,13 +130,12 @@ import Web.Scotty
       literal,
       post,
       raw,
-      scotty,
-      scottyApp,
-      ScottyM,
+      Options(..),
       setHeader,
       status,
       text )
-import Network.Wai.Handler.Warp (Port, defaultSettings, setPort)
+
+import Network.Wai.Handler.Warp (Port, defaultSettings, setPort, setFork)
 import Network.Wai.Handler.WarpTLS (runTLS, tlsSettings)
 
 import qualified Argo.Doc as Doc
@@ -459,6 +464,8 @@ interruptAllThreads = Notification $ do
 
 --------------------------------------------------------------------------------
 
+newtype ThreadQueue = ThreadQueue {threadQueueMVar :: MVar (IORef (Seq ThreadId))}
+
 -- | An application is a state and a mapping from names to methods.
 data App s =
   App
@@ -468,9 +475,8 @@ data App s =
   , appDocumentation :: [Doc.Block]
   , appActiveThreads :: !(MVar (IORef (Set ThreadId)))
   -- ^ Thread currently active processing requests or waiting to do so.
-  , appRequestLock :: MVar ()
-  -- ^ Lock used to pause new request handling when needed and prevent
-  -- multiple notifications from executing concurrently.
+  , appThreadQueue :: ThreadQueue -- !(MVar (IORef (Natural, Set (Natural, ThreadId), Map ThreadId Natural)))
+  -- ^ Used to create an ordering on threads to ensure in-order servicing of requests.
   }
 
 
@@ -487,15 +493,15 @@ mkApp ::
 mkApp name docs opts initAppState methods = do
   initialState <- newMVar =<< initServerState (appStateMutability opts) initAppState
   initThreadPool <- newMVar =<< newIORef Set.empty
-  reqLock <- newMVar ()
+  threadQueue <- ThreadQueue <$> (newMVar =<< newIORef Seq.empty)
   pure $ App
     { serverState = initialState
-    , appMethods = M.fromList [ (methodName m, m)
-                              | m <- methods]
+    , appMethods = Map.fromList [ (methodName m, m)
+                                | m <- methods]
     , appName = name
     , appDocumentation = appDocs
     , appActiveThreads = initThreadPool
-    , appRequestLock = reqLock
+    , appThreadQueue = threadQueue
     }
   where appDocs =
           docs ++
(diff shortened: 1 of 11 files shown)


pnwamk:
This set of changes has two main consequences:
1. cryptol/saw python scripts  should not have to reset the server each time they connect; new connections' states take precedence over older connections' states and the older states are auto-evicted when the server hits it's capacity. The capacity can be customized from the command line (`--max-occupancy NUM`) and the old behavior (i.e., not auto evicting) can be regained from the command line as well (`--no-evict`)
2. the "unknown state ID" errors should be a thing of the past because (a) now that we can skip resetting the server (which was the request occasionally "colliding" with the subsequent request on the server) and (b) several measures have been taken to ensure that even if people send notifications (such as a reset server command) they should not be interrupted and/or conflict with requests immediately following them. I have tested this locally (i.e., running a program that reliably caused the error previously in a loop) with the cryptol server and even with the `reset_server=True` flag (which we no longer need) no unknown state errors occurred.

argo/src/Argo/DefaultMain.hs
```diff
@@ -284,14 +287,20 @@ maxOccupancyOpt =
    (Opt.long "max-occupancy" <>
     Opt.help "Maximum number of active sessions allowed at once." <>
     Opt.showDefault <>
-    Opt.value 10))
+    Opt.value 1))
```
m-yac, line 290: Does this mean that a server only saves one previous state now? Or if we go by the description immediately above, does this mean that a server can only connect to a single client at a time now? Neither of these seem to me like something we want.

argo/src/Argo/DefaultMain.hs
```diff
@@ -284,14 +287,20 @@ maxOccupancyOpt =
    (Opt.long "max-occupancy" <>
     Opt.help "Maximum number of active sessions allowed at once." <>
```
m-yac, line 288: Based on other comments about this option in this repo and in your PR, shouldn't this be "Maximum number of states allowed at once"? That seems like an important difference.

argo/src/Argo.hs
```diff
@@ -851,15 +864,17 @@ handleRequest ::
...
     Just (CommandMethod m) -> withActiveThread app $ withRequestID $ \reqID -> do
       stateID <- getStateID req
       response <- withMVar theState $ \state -> do
-        when (stateID == initialStateID) $ do
-          stateCount <- statePoolCount state
-          when (stateCount >= (optMaxOccupancy opts)) $
-            throwIO serverAtCapacity
+        serverFull <- do cnt <- statePoolCount state
+                         pure $ cnt >= (optMaxOccupancy opts)
+                                && stateID == initialStateID
```
m-yac, line 875: Why do we only consider the server full if the request's state ID is the initial state ID? Wouldn't this mean that the server would only ever be considered full if `optMaxOccupancy == 1`?

m-yac (commented):
This looks awesome! I have a few questions, but they could just be due to me not having a full understanding of this codebase yet.

argo/src/Argo.hs
```diff
@@ -888,7 +905,9 @@ handleRequest opts respond app req = do
               result <- execQuery (q params) reqID qctx appState
               return $ addStateID stateID result
       respond (JSON.encode response)
-    Just (NotificationMethod m) ->
+    Just (NotificationMethod m) -> do
+      -- Acquire the request lock so _new_ requests cannot be served
+      -- while the notification is executing.
```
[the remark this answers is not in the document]
pnwamk: Aha! No they are not. Thanks.

argo/src/Argo/DefaultMain.hs
pnwamk: Yes, it essentially means by _default_ the server will connect to at most one client simultaneously (a new client connection which initiates some computation would cause the previous client's state on the server to be evicted). It can be customized at the command line (there's some tests that do that, e.g., upping it to `2` and then having their connections interact via interrupts etc).

I chose `1` as I believe this is the most common (only?) use case in practice from what I've observed thus far, and it will keep the server's memory footprint small for that use case. Happy to change it if we think it's a problem.

It might also be nice at some point to make this a dynamic server option as well... but again, it's hard to know based on our limited observed use cases. The impression I get is the server is being used exclusively for single connections performing linear series of computations, so I defaulted it to be optimized for that with knobs for other desires.

argo/src/Argo/DefaultMain.hs
pnwamk: Technically from an internal view I would say you're 100% right. Practically though, from a user point of view, you never see the "states" (they're implementation details of the connection/interactions), so I'm biased towards a description more phrased like how it is now. As far as a user/connection is concerned, it _is_ the number of active sessions allowed I think it's fair to say.

I'm happy to rephrase it -- I can't think off the top of my head on a different wording that avoids too much low level detail but still speaks to the observable behavior for clients.

Ah, but even if we leave the documentation string as is... it's probably worth adding a comment near that in the Haskell code to clarify for a developer reading that! That is confusing without all that other history/context/etc.

argo/src/Argo.hs
pnwamk: If a command is made from some non-initial state `X` and it creates some new non-initial state `Y`, `X` is destroyed and replaced with `Y` (as things are currently implemented -- historically this has not always been the case). So the number of live states is maintained. We only get an _additional_ state if it's a fresh request, since we always keep around the initial state.

I'll add a quick note to clarify.

argo/src/Argo.hs
```diff
@@ -935,8 +956,11 @@ handleRequest opts respond app req = do
 interruptAllAppThreads :: App s -> IO ()
 interruptAllAppThreads app = do
   withMVar (appActiveThreads app) $ \threadsRef -> do
-    mapM_ killThread =<< readIORef threadsRef
+    activeThreads <- readIORef threadsRef
+    mapM_ killThread activeThreads
     writeIORef threadsRef Set.empty
+    withMVar (threadQueueMVar (appThreadQueue app)) $ \qRef -> do
+      modifyIORef' qRef (Seq.filter (flip Set.member activeThreads))
```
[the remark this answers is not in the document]
pnwamk: You are exactly right! My brain was thinking "delete them" and forget to invert the logic for what filtering's predicate does! Doh!

argo/src/Argo/DefaultMain.hs
atomb: Yeah, I think it's reasonable to have the default be `1`.

python/argo_client/connection.py
```diff
@@ -472,6 +473,11 @@ def send_notification(self, method: str, params: dict) -> None:
                'params': params}
         msg_string = json.dumps(msg)
         self.process.send_one_message(msg_string, expecting_response = False)
+        # Since notifications require no response and can interrupt commands/queries,
+        # they can sometimes interrupt subsequent requests. We have tried to guarantee
+        # in-order servicing on the server. This short pause is an additional small
+        # safety measure to keep things running smoothly.
+        time.sleep(0.1)
```
atomb, line 480: This is somewhat unfortunate, and seems like it's fragile in the long run. I'm happy to include it for now, to get things working, but I'd hope that we can figure out a way to not need it in the long run.

atomb (approved):
This looks good to me. I hope we don't need to keep the delay in the Python client for long, and can figure out how to remove the need on the server side, but I think it's okay to include for now.

python/argo_client/connection.py
pnwamk: I just went back and tested this a final time, making sure the cryptol server was in fact using the _new_ argo updates in this PR, and the explicit pause is no longer necessary to avoid the unknown state id errors even in the pathological looping case. I'll confirm again tomorrow when I do the cryptol bump, but I'm pretty sure we're good-to-go now. (Last time I must have updated the client but not build and _reinstalled_ the server, doh.)

PR #193, opened 2022-11-28
Relax upper bounds to support GHC 9.4.3

### LINKED ISSUE #192: Bump cabal boundaries to support GHC 9.4.3

yav:
Fixes #192

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
     - uses: actions/setup-python@v2
       with:
-        python-version: '3.x'
+        python-version: '3.7'
     - uses: abatilo/actions-poetry@v2.1.4
       with:
         poetry-version: 1.1.11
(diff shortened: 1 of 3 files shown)


RyanGlScott:
To fix the Python installation error, I'd recommend changing these lines:

https://github.com/GaloisInc/argo/blob/7433f90d9fdf81efe5916bc9c66cf28439efb0ec/.github/workflows/ci.yml#L19-L21

To something like:

```yaml
      - uses: actions/setup-python@v2
        with:
          python-version: '3.7'
```

(I took inspiration from the [corresponding lines](https://github.com/GaloisInc/cryptol/blob/bf17c70ae3ae0f5e0393efb9c52658dac9016694/.github/workflows/ci.yml#L88-L90) in Cryptol's `ci.yml`, which are known to work.)

RyanGlScott (approved):
It would be worth adding `"9.4.3"` to the CI matrix here to ensure that it actually builds with GHC 9.4:

https://github.com/GaloisInc/argo/blob/7433f90d9fdf81efe5916bc9c66cf28439efb0ec/.github/workflows/ci.yml#L14

Otherwise, LGTM.

PR #195, opened 2023-01-26
Python: Avoid `typed-ast` dependency

RyanGlScott:
The `typed-ast` package is being end-of-life'd soon. See python/typed_ast#179. Moreover, it is causing issues when building with Python 3.11+. See GaloisInc/cryptol#1491.  Thankfully, `typed-ast` is no longer required if you use sufficiently new versions of `mypy` and Python. This patch raises the lower version bounds to make it possible to shed the `typed-ast` dependency entirely.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
     - uses: actions/setup-python@v2
       with:
-        python-version: '3.7'
+        python-version: '3.11'
     - uses: abatilo/actions-poetry@v2.1.4
       with:
         poetry-version: 1.1.11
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         poetry install
     - name: Typecheck python code
       working-directory: ./python
-      run: poetry run mypy argo_client
+      run: poetry run mypy --install-types --non-interactive argo_client
```
RyanGlScott, line 54: This change is a bit unfortunate, but perhaps unavoidable. `mypy-0.900` do not install third-party libraries' type packages unless you explicitly opt in to it with [`--install-types`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-install-types). Moreover, I'm unclear if it's possible to make `mypy` always use `--install-types` as a `pyproject.toml` configuration option. At the very least, I can't find anything about this in the [list of configuration options](https://mypy.readthedocs.io/en/stable/config_file.html).

kquick (approved):

PR #205, opened 2024-01-18
Allow building with GHC 9.8

RyanGlScott:
This bumps the upper version bounds on the `base`, `bytestring`, `scotty`, `text`, and `vector` libraries in order to support building with versions that are required by GHC 9.8.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ghc-ver: ["8.8.4", "8.10.7", "9.0.2", "9.2.8", "9.4.5", "9.6.2"]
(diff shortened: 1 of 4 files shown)


RyanGlScott:
I believe the test failures are due to `argo` using a more recent build plan that includes `aeson-2.2.*`, which changed some of its error messages (see [this commit](https://github.com/haskell/aeson/commit/e79a7ca4515ad9c9b62e7da989ec05e4a1013240)). I'll have to think about whether we want to raise the lower version bounds on `aeson`, accept both forms of error messages, or something else.

kquick:
I think all of the test failures come from https://github.com/GaloisInc/argo/blob/master/python/tests/test_file_echo_api.py#L197 and this could be simply modified to ensure that the `invalid_request` value is contained in the `data.data` JSON element, then remove that element and verify equivalency of the result modulo that element.

In fact, I think we are better off now, considering the original input and output differed by an internal space character!

kquick (approved):

PR #209, opened 2024-05-15
Bump argo to python 3.12

mccleeary-galois:
Bump argo to support python 3.12


python/mypy.ini
```diff
@@ -1,6 +1,6 @@
 [mypy]
 no_implicit_optional = True
-python_version = 3.7
+python_version = 3.12
```
RyanGlScott, line 3: Worth bumping the Python version in the CI as well ([here](https://github.com/GaloisInc/argo/blob/59823e8fd6985986abd9ae784932b94bd735b17d/.github/workflows/ci.yml#L22)). Make sure to also leave a comment here pointing to this part of the CI configuration (and vice versa) so that if we bump the Python version in the future, we remember to update both locations.

python/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
++ Bump `mypy` to  `mypy-1.10`, and update its dependencies to support the bump. This allows for Python 3.12 support.
```
RyanGlScott, line 1: I like to include a header for not-yet-released versions. Something like:

```suggestion
# argo-client next (TBA)
+ Bump `mypy` to  `mypy-1.10`, and update its dependencies to support the bump. This allows for Python 3.12 support.
```

RyanGlScott (approved):

PR #210, opened 2024-05-15
Setup for release of v0.0.12

mccleeary-galois:


python/mypy.ini
```diff
@@ -1,5 +1,6 @@
 [mypy]
 no_implicit_optional = True
+# Ensure CI file matches python version here.
```
RyanGlScott, line 3: Specifically:

```suggestion
# Ensure .github/workflows/ci.yml matches the Python version used here.
```

.github/workflows/ci.yml
```diff
@@ -19,6 +19,7 @@ jobs:
     steps:
     - uses: actions/setup-python@v2
       with:
+        # Ensure TOML file and mypy.ini file are updated to support version here.
```
RyanGlScott, line 22: Specifically:

```suggestion
        # Ensure pyproject.toml and python/mypy.ini are kept in sync with the Python version here.
```

RyanGlScott (approved):
Thanks! One additional thing: can you change the commit message from "[Update version numbers for new release" to something that mentions 0.1.12 in particular? That way, it's easier to tell from the `git log` what version it's about.

PR #216, opened 2024-09-04
Using blocking network IO to avoid busy-wait

glguy:
This change also makes the netstring parser more robust to support the way that it's used in the network loop

### DIFF
--- a/python/argo_client/connection.py
+++ b/python/argo_client/connection.py
@@ -158,32 +158,22 @@ class SocketProcess(ManagedProcess):
         self.persist = persist
         super().__init__(command, environment=environment)
 
-
-    def buffer_replies(self) -> None:
-        """Read any replies that the server has sent, and add their byte
-           representation to the internal buffer, freeing up space in
-           the pipe or socket.
-        """
-        try:
-            arrived = self.socket.recv(4096)
-            while arrived != b'':
-                self.buf.extend(arrived)
-                arrived = self.socket.recv(4096)
-            return None
-        except BlockingIOError:
-            return None
-
     def get_one_reply(self) -> Optional[str]:
-        """If a complete reply has been buffered, parse it from the buffer and
-           return it as a bytestring."""
-        self.buffer_replies()
-        try:
-            (msg, rest) = netstring.decode(self.buf)
-            self.buf = bytearray(rest)
-            self._log_rx(msg)
-            return msg
-        except (ValueError, IndexError):
-            return None
+        """Return the next message if there is one. Block until the message
+        is ready or the socket has closed. Return None if the socket closes
+        and there are no buffered messages ready."""
+        while True:
+            got = netstring.decode(self.buf)
+            if got is None:
+                arrived = self.socket.recv(4096)
+                if arrived == '':
+                    return None
+                self.buf.extend(arrived)
+            else:
+                (msg, rest) = got
+                self.buf = bytearray(rest)
+                self._log_rx(msg)
+                return msg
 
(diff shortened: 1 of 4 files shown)


python/argo_client/netstring.py
```diff
@@ -13,7 +13,13 @@ def encode(string : str) -> bytes:
...
-def decode(netstring : bytes) -> Tuple[str, bytes]:
+class InvalidNetstring(Exception):
+    """Exception for malformed netstrings"""
+    def __init__(self, message: str):
+        self.message = message
+        super().__init__(self.message)
+
+def decode(netstring : bytes) -> Optional[Tuple[str, bytes]]:
     """Decode the first valid netstring from a bytestring, returning its
     string contents and the remainder of the bytestring.
```
RyanGlScott, line 24: Please document under what conditions this can return `None`.

python/argo_client/netstring.py
```diff
@@ -22,20 +28,26 @@ def decode(netstring : bytes) -> Tuple[str, bytes]:
...
-        out.append(netstring[i])
-        i += 1
-    if chr(netstring[i]).encode() != b',':
-        raise ValueError("Malformed netstring, missing ,")
-    i += 1
-    return (out.decode(), netstring[i:])
+    colon = netstring.find(b':')
+    if colon == -1 and len(netstring) >= 10 or colon >= 10:
+        # cut things off at about a gigabyte
+        raise InvalidNetstring("message length too long")
```
RyanGlScott, line 34: I think I'm missing something: how does this code check if the netstring is over a gigabyte? (The number `10` would suggest a much smaller amount.)

python/argo_client/connection.py
```diff
@@ -158,32 +158,22 @@ def __init__(self, command: str, *,
...
-           the pipe or socket.
-        """
-        try:
-            arrived = self.socket.recv(4096)
-            while arrived != b'':
-                self.buf.extend(arrived)
-                arrived = self.socket.recv(4096)
-            return None
-        except BlockingIOError:
-            return None
```
RyanGlScott, line 174: Since we deleted a method from this class, this PR technically changes the `argo-client` API. It would be worth documenting this change in the changelog (as well as the removal of `RemoteSocketProcess.buffer_replies` and `ServerConnection._process_replies`).

(For what it's worth, I checked and neither `cryptol` nor `saw-client` appear to use any of these methods directly, so I doubt there will be any breakage in practice.)

RyanGlScott (commented):
Have you verified that the test suite for the Cryptol Python bindings continues to pass after these changes? (I recall that getting the network code just right for the test suite was very tricky, so it would be good to check to be on the safe side.)

python/argo_client/netstring.py
glguy: a gigabyte is 1073741824 which is ten digits long. if we're getting more than 10 digits then something is wrong and I wanted to detect that early. I wasn't really intending for this to be a meaningful size restriction. I can do that if it's important

RyanGlScott: Ah, I thought you meant gigabytes in the sense of the _length_ of the netstring, but you meant gigabytes in the _value_ of the netstring. That clarifies my confusion (although it would be worth adding a similar clarification in the comment as well).

glguy:
> Have you verified that the test suite for the Cryptol Python bindings continues to pass after these changes? (I recall that getting the network code just right for the test suite was very tricky, so it would be good to check to be on the safe side.)

./run_rpc_tests.sh passes

Issue #218, opened 2025-03-03
`Argo.DefaultMain`: Allow some degree of global option configuration (e.g., to enable a `--version` option)


RyanGlScott:
While `argo`'s `Argo.DefaultMain.defaultMain` function provides a default command-line argument parser that is suitable for most applications, there are corner cases where it is convenient to supplement the default parser using [`Argo.DefaultMain.customMain`](https://github.com/GaloisInc/argo/blob/5a56a26a32c1d7216fd0bb5ef70209f4f5d1f011/argo/src/Argo/DefaultMain.hs#L58-L74). `customMain` suffers from a serious limitation, however: it only allows you to add command-line options for the application's _subcommands_ (`<APP> stdio`, `<APP> socket`, `<APP> http`, and `<APP> doc`). It does not allow you to add add add command-line options for the top-level command, which makes it difficult to support common options like `<APP> --version`. (See https://github.com/GaloisInc/cryptol/issues/1815 and https://github.com/GaloisInc/saw-script/issues/2237 for use cases.)

I propose that we extend `customMain` to make it possible to, at a minimum, implement `--version`. This may require a backwards-incompatible change to `customMain` to accomplish, but perhaps we can avoid changing `defaultMain` (what most `argo`-based applications use) as well. Here are some various options that come to mind:

1. Don't change the type of `customMain` at all, but instead require the user to write `<APP> <SUBCOMMAND> --version` instead of `<APP> --version` (e.g., `<APP> stdio --version`). Then, the user could instantiate the four pre-existing `Parser` arguments to `customMain` to take a `--version` option.

   This has the advantage of not needing to change anything in `argo`, but it comes with the drawback that the user must remember to supply a subcommand when querying the version. This is quite counterintuitive.
2. Add an additional argument of type `(forall a. Parser (a -> a))` to `customMain` to declare top-level options. For instance, an application could instantiate this argument with `simplerVersioner <VERSION NUMBER>` (using `optparse-applicative`'s [`simplerVersioner`](https://hackage.haskell.org/package/optparse-applicative-0.18.1.0/docs/Options-Applicative.html#v:simpleVersioner) combinator) to add a `--version` option that prints `<VERSION NUMBER>` when selected. `defaultMain` could simply instantiate this argument with `pure id` as a default.

[...]

PR #220, opened 2025-03-04
Add `customMain` argument for `--version`-like global options

RyanGlScott:
`argo`'s allows users to customize the CLI option-parsing logic via the `customMain` function. The primary means to do so is by adding options for each subcommand (e.g., `stdio`). However, it is also convenient to allow users to specify global options for things that look like a `--version` option—i.e., that print some information and exit.

This patch enables this functionality by adding an extra `(forall a. Parser (a -> a))` argument to `customMain`, which can be used to configure global options that resemble `--version`. I have also updated the `file-echo-api` example applications in this repo to support a `--version` option, which provides a demonstration of how this functionality can be used.

Fixes https://github.com/GaloisInc/argo/issues/218.

### DIFF
--- a/README-dist.rst
+++ b/README-dist.rst
@@ -8,7 +8,7 @@ convenient interaction with that API.
 Environment Setup
 =================
 
-The package contains two directories that external tools need ot be able
+The package contains two directories that external tools need to be able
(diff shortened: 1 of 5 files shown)


sauclovian-g (approved):
LGTM

PR #224, opened 2025-11-06
Prepare release 0.0.14

aschwerdfeger-galois:


RyanGlScott (approved):
Thanks!

Now that I check, we've not been doing a very good job of keeping the [`argo` changelog](https://github.com/GaloisInc/argo/blob/master/argo/CHANGELOG.md) up to date. This isn't really your fault, though—we would need to go back through `argo`'s release history and add the missing changelog entries at some point.

Issue #227, opened 2026-01-08
CI failure: Python tests fail due to `ConnectionResetError: [Errno 104] Connection reset by peer`


RyanGlScott:
Recent CI jobs, such as [this job from the `master` branch](https://github.com/GaloisInc/argo/actions/runs/20051027061/job/59791695997), have exhibited several failures in the Python tests. All of the failures look roughly like this:

```
======================================================================
ERROR: test_basics (test_file_echo_api.HttpTests.test_basics)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/.cache/pypoetry/virtualenvs/argo-client-eMHzIW9q-py3.12/lib/python3.12/site-packages/urllib3/connectionpool.py", line 787, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
  File "/home/runner/.cache/pypoetry/virtualenvs/argo-client-eMHzIW9q-py3.12/lib/python3.12/site-packages/urllib3/connectionpool.py", line 534, in _make_request
    response = conn.getresponse()
               ^^^^^^^^^^^^^^^^^^
  File "/home/runner/.cache/pypoetry/virtualenvs/argo-client-eMHzIW9q-py3.12/lib/python3.12/site-packages/urllib3/connection.py", line 571, in getresponse
    httplib_response = super().getresponse()
                       ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/http/client.py", line 1430, in getresponse
    response.begin()
  File "/opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/http/client.py", line 331, in begin
    version, status, reason = self._read_status()
                              ^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/http/client.py", line 292, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/socket.py", line 720, in readinto
    return self._sock.recv_into(b)
           ^^^^^^^^^^^^^^^^^^^^^^^
ConnectionResetError: [Errno 104] Connection reset by peer
```

[...]

PR #230, opened 2026-01-09
Compile examples with `-threaded`, allow `time-manager-0.3.*`

RyanGlScott:
This is a more robust fix for #227 that allows `argo` to build against `time-manager-0.3.*` instead of restricting the build to old `time-manager` versions. The new time manager implementation in `time-manager-0.3.*` currently requires `-threaded` (see https://github.com/yesodweb/wai/issues/1053#issuecomment-3728839511), so make sure that the executables in `file-echo-api.cabal` use this flag.


sauclovian-g (commented):
LGTM... however, don't you also want to remove the time-manager constraint?

PR #231, opened 2026-01-15
Allow building with GHC 9.12 (and 9.10)

sauclovian-g:
Bump upper version bounds on base, containers, and filepath to include the versions in those releases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


RyanGlScott (commented):
Can you also add CI jobs for GHC 9.10 and 9.12?

sauclovian-g:
This failure with 9.0.2 is apparently repeatable; on the other hand, it's also inexplicable, so I'm inclined to blame it on cache corruption. Any countervailing opinions?

sauclovian-g:
Guess so.

aschwerdfeger-galois (approved):

